PR #1, opened 2025-11-08
Scope-crept loc/item names

flibber-hk:
These correspond to the ones I extracted, plus Mr Mushroom, minus bonetown top (which doesn't seem to exist). I'm going to write an essay in discord about the trobbio/quill location stuff.

Technical details: This is all the BasicNPCs I could find, filtered for things that are NPCs (rather than tablets), things that don't seem to exist, and things I don't consider to be lore tablets (like the sign on Pinsmith's door when he goes to shellwood). I skimmed the list of PlayMakerNPCs and could only spot the Mr Mushroom one, I doubt there are any others but there might be.


flibber-hk:
Note - please review before merging, there are some things that are open to interpretation that probably need opinions (anything worth mentioning has been described in the comments)

flibber-hk:
Slight scope creep

PR #9, opened 2026-01-04
Mask Shards, Spool Fragments, Pale Oil & Silk Hearts

nerthul11:


homothetyhk:
"Mask_Shard-Savage_Brightvein" should be "Mask_Shard-Brightvein". I'm ok to leave the comments in and revisit names later before release if people feel like they are confusing/misleading.

PR #11, opened 2026-01-15
Changed Lore Tablet Names and added new lore tablets that weremissing

DatUub:


DatUub:
Why is that still enabled :/

Issue #17, opened 2026-01-24
Add docs and workflow for publishing docs


BadMagic100:
Using docfx. A skeleton is fine, at the minimum we just need the auto-generated API stuff, articles may come later

PR #36, opened 2026-02-03
Docs workflow 

### LINKED ISSUE #15: Add publishing actions and other template upgrades
This was created on a very early version of the silksong template which means it's missing a number of things one would like to have; this is to pull in those changes to the csproj and add in the actions workflow

DatUub:
Set up build automation and the documentation framework. 
Closes #15 & #17.

### Changes
* **CI/CD:** Builds on PRs and auto-releases on push to master.
* **Docs:** Configured DocFX to deploy to GitHub Pages on release.

### Questions
- [ ] **Thunderstore:** Since this is a library, should we publish there?
- [ ] **NuGet:** Should we add NuGet publishing to the workflow?
- [ ] **Analyzers:** Should we enable standard .NET analyzers in the project file?


BadMagic100:
The answer to all of the 3 questions is yes (similarly to the workflow in the template publishing should be suppressed by default)
* Thunderstore - It needs to be published so that mods can take a dependency on it and have it installed, it'll have to vendor ItemChanger.Core (I believe flib already did this though)
* NuGet - Taking dependencies via nuget is preferred so that other mods can use it in workflows easily
* Analyzers - they're just good (mostly)

BadMagic100 (changes requested):

.github/workflows/github-pages.yml
```diff
@@ -29,18 +29,15 @@ jobs:
...
+          dotnet-version: '10.x'
 
       - name: Build Project
         run: dotnet build ItemChanger.Silksong/ItemChanger.Silksong.csproj -c Release
         continue-on-error: true
 
       - name: DocFX build
         working-directory: docs
-        run: docfx docfx.json
+        run: dnx docfx.json
```
BadMagic100: this is not correct, I suggest actually reading the docs about the tools you're being suggested to use before using them, and then testing that it actually works

DatUub: I swear i edited it so it would be dnx 😭

BadMagic100 (changes requested):

homothetyhk:
With #15, my expectation was that the resulting actions and thunderstore directory structure would be identical to that proposed by the template (e.g. https://github.com/homothetyhk/Silksong.Benchwarp/commit/5336e21b2dd7747fdff35bd5b74b40bdb3f3cd70#diff-dbd515845a90a91f83cd2dbcdf41150bff6a75806900384a44afb1f0650d9cc9). I care about the project adhering to the community standard more than anything else.

docs/api/.gitignore
BadMagic100: now that I look at this again it's probably more idiomatic to just ignore `api/` in the `docs`-level gitignore... but if this is being kept it should have a `**/*` to actually ignore the contents

.github/workflows/github-pages.yml
```diff
@@ -0,0 +1,52 @@
+name: Docs
+
+on:
+  release:
+    types: [published]
```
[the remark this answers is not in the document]
BadMagic100: I did a brief search and it seems possible but extremely nontrivial to have docs for multiple/historical versions, to the point that it's probably not ever worth the investment.

Unrelated to the above revelation I now think it's much better to trigger on push, especially since the first release is a way off

PR #45, opened 2026-02-05
added RelicUIDef and ToolUIDef classes

KSPlayer1337:
initial classes for RelicUIDef and ToolUIDef


KSPlayer1337:
should fix this issue https://github.com/homothetyhk/ItemChanger.Silksong/issues/46

flibber-hk:
Maybe this is one for Homothety to mull over, but I think there’s a lot of duplicated code here and it’d probably be better to consolidate these classes along with the existing collectable item one into a single SavedItemUIDef that wraps an ISavedItem (with implementations for tools, relics, etc)

Issue #24, opened 2026-01-31
Persist profile with DataManager


dpinela:
Ideally this should (transparently) store immutable parts of the profile using OnceSaveData and only mutable ones with mutable SaveData.

BadMagic100:
I think that
1. Separating immutable data, while a good idea, is blocked on https://github.com/BadMagic100/ItemChanger.Core/issues/13 and is not critically important for launch
2. Due to IC's persistence APIs, I think as-is you'd have to roll your own by writing files to the right paths. If there's a sensible API that can be added to get all the custom converters into the right places to pass objects to datamanager I'd be open to an IC.core PR that does this

dpinela:
I have no problem with delaying immutable data provided we can introduce it later without breaking clients of IC, also preferrably without breaking save compatibility (though historically we've advised against updating mods mid-save anyway).

For the converters, does the JsonConverter attribute not cover every case where we'd want to attach a converter to a type we want to serialize? We can put it on both types and properties.

BadMagic100:
I'm pretty sure I had a good reason to force serialization to roundtrip through ic.core's APIs rather than making it easy to diy it, I'll go look at the related code and see if I can remember why. Off the top of my head I can think of reasons that the API should exist but not justify why no alternative

dpinela:
I tested making it go through DataManager anyway and it worked as expected (at least with ItemChangerTesting and StartCrestSelector, which are all the clients that exist right now?). What was the reason you had to force serialization to go through IC.Core?

In my opinion, IC.Core _should_ offer a supported way of doing this, if nothing else because otherwise people who need/want it will use _unsupported_ ones (like I did).

BadMagic100:
These converters also matter somewhat, particularly I believe that between the two of these it is probably cumbersome to tag every usage/remember that they should be tagged

https://github.com/BadMagic100/ItemChanger.Core/blob/main/ItemChanger.Core/Serialization/SerializationHelper.cs#L28

dpinela:
> Ah I forgot about this, but I believe I've defined the APIs to enforce that this happens on profile load

This much I understood; what I thought you were going to explain is _why_ you defined the API to enforce that in the first place. Because that's what is blocking this.

BadMagic100:
> > Ah I forgot about this, but I believe I've defined the APIs to enforce that this happens on profile load
> 
[...]

PR #47, opened 2026-02-07
Persist profile with DataManager

dpinela:
This is not intended to merge as-is; the private method invocations should be replaced by some (TBD) proper API from IC.Core.


dpinela:
This implements #24

dpinela:
With the new API in DataManager, this is ready to merge now.

Issue #21, opened 2026-01-29
Flea container


flibber-hk:
I'm making this issue to indicate that I'm working on it, but I don't want to commit to the whole thing so I may end up making a partial PR and update this issue with further tasks.

In scope:

* Flea container
* Flea item(s), UIDef and Postview IString
* Flea locations

Out of scope:

* Very special flea locations (Vog, Kratt, Bigflea)
* Vog shop
* Mooshka shop
* Flea transport (I will PR some of my work on this later / separately, depending on how much should get added as part of base rando)

flibber-hk:
This is now blocked by https://github.com/BadMagic100/ItemChanger.Core/issues/27 :zotesip:

PR #48, opened 2026-02-08
Flea container

flibber-hk:
The scope for this PR is anything associated with the flea container, so that has become:

* The container
* The locations
* The item
* The UI Def

As discussed on discord, this involved updating IC.Core to v0.5.0, and I have modified the codebase to comply with the breaking changes. (The changes to container capabilites are not reflected here because I'm considering capabilities to be out of scope)

----

There are a few TODOs that I haven't completed, but I've signposted them. This is a mixture of me limiting scope and laziness.

* Set up the container capabilities (as discussed). This may involve different capabilities for each instantiable flea (for example the sleeping flea might not support drop, the slab flea might have a special capability for being able to be instantiated on the ceiling, etc)
* Consider making the flea container fling a shiny or shinies for non-flea items, or fling rosaries if that makes sense, etc
* Write something funny for the shopdesc - I just put in a placeholder
* Make all flea locations replaceable - this would involve measuring the elevation for each location other than the ones I've already done (which are the instantiable ones plus the slab cell one).
* Move profile instantiation to Awake.

There are a few more (such as figuring out how to handle the special fleas) but that's far beyond the scope of this PR.

----

Some notes:

* I feel that having anonymous fleas is much better than declaring that flea items are associated with an arbitrary scene. I believe everything in the game that checks fleas saved count goes via either SavedFleasCount or GetVariables, and all the important stuff goes through the saved item (which goes via GetVariables and adds on checks for the special fleas). The alternative would be to declare that e.g. there's a flea item that sets SavedFlea_Ant_03 and a flea item that sets SavedFlea_Slab_Cell and they differ in that each says it prefers a different flea container subtype.
* I wanted to introduce a way to make consistent random selections for a flea container, I'm pretty happy with the general approach I came up with (the idea being that this can be reused for things like rainbow eggs) although the actual randomization func I used may be able to be improved. The seed is randomly selected by default but should be derived from the rando seed in rando (obviously).
* I don't know where the best place to put the fleas collected saved item (as loaded by assethelper) - maybe you want to set up an object cache or something, it needs to be able to be loaded while in a profile though which is why I put it on the container without really thinking.
* Elevations (yes I know they're called Correction in IC.Core but in the IC.SS context I think elevation is fine). I believe it is the responsibility of IC.Silksong to document a consistent strategy for computing the elevation of a container and of an object location. My approach has been:
  - The elevation of an object location is uniformly (asset y - hornet y), where hornet y is measured where you expect to pick up the shiny
  - The elevation of a container (the value passed to WithLocalOffset) is uniformly (asset y - hornet y), measured in the scene the asset is loaded from

This might mean that elevations need to be recalculated for existing object locations (or maybe it doesn't matter), but I had to change the y coordinate of the Tut_02 shiny to reflect this (I didn't really test the whispering vaults location thoroughly but I think it works)


ItemChanger.Silksong/Modules/AnonymousFleasModule.cs
```diff
@@ -0,0 +1,95 @@
...
+/// <summary>
+/// Module recording the number of fleas saved that are not associated to a particular scene.
+/// </summary>
+[SingletonModule]
+public sealed class AnonymousFleasModule : Module
+{
+    /// <summary>
+    /// The number of anonymous fleas available in the file.
+    /// </summary>
+    [JsonIgnore]  // Ignore from json because it's set by the flea items as they load
```
homothetyhk: @BadMagic100 Is this compatible with the late item loading in AP?

BadMagic100: No, this'll be problematic. AP gives no reasonable way to know about item placements in other worlds so this count will be off without special treatment. The best I can probably do without getting in trouble for overusing slot data is grab the expected total, but even this clashes with the behavior in FleaItem.DoLoad, so I think it this will need further design consideration

ItemChanger.Silksong/Modules/ConsistentRandomnessModule.cs
BadMagic100: My personal expectation is that variant selection in IC is obviously deterministic, possibly with a subset of valid variants randomly chosen by rando in the request builder. Having the randomness here is mixing responsibilities imo

ItemChanger.Silksong/ItemChanger.Silksong.csproj
```diff
@@ -53,7 +53,9 @@
     <PackageReference Include="UnityEngine.Modules" Version="6000.0.50" IncludeAssets="compile" />
     <!--If you're unable resolve this dependency, check that you're using the nuget v3 package feed instead of v2.-->
     <PackageReference Include="Silksong.GameLibs" Version="1.1.0-silksong1.0.28650" />
-    <PackageReference Include="ItemChanger.Core" Version="0.4.1" GeneratePathProperty="true" />
+    <PackageReference Include="Silksong.AssetHelper" Version="1.2.0" />
+    <PackageReference Include="Silksong.FsmUtil" Version="0.3.12" />
+    <PackageReference Include="ItemChanger.Core" Version="0.5.0" GeneratePathProperty="true" />
```
BadMagic100: Probably want to go to 0.5.2 for tag constraint enforcement

ItemChanger.Silksong/RawData/BaseItemList.cs
```diff
@@ -15,6 +15,19 @@ internal static class BaseItemList
             UIDef = new CollectableUIDef { CollectableName = "Ward Boss Key" },
         };
 
+        public static Item Flea => new FleaItem
+        {
+            Name = ItemNames.Flea,
+            UIDef = new MsgUIDef()
+            {
+                // TODO - improve the shopdesc
```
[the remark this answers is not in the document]
flibber-hk: I’m going to assume this comment is a joke

ItemChanger.Silksong/Tags/DeactivateIfContainerChangedTag.cs
```diff
@@ -0,0 +1,49 @@
...
+/// Deactivate the listed object if the location's container does not match the expected container.
+/// </summary>
+[TagConstrainedTo<ContainerLocation>]
+public class DeactivateIfContainerChangedTag : Tag
+{
+    public required string SceneName { get; init; }
+
+    public required string ObjectName { get; init; }
+
+    public required string ExpectedContainerType { get; init; }
```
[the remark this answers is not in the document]
flibber-hk: Yeah, I generalized the function without changing the name :zote:

ItemChanger.Silksong/FsmStateActions/PlacementAllObtainedCheck.cs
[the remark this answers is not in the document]
flibber-hk: Probably yes

ItemChanger.Silksong/Modules/AnonymousFleasModule.cs
flibber-hk: I’m not particularly attached to the minutiae so I’d be happy for someone else to figure out the details if they’re opinionated (I *am* attached to the notion of anonymous fleas though)

ItemChanger.Silksong/ItemChanger.Silksong.csproj
flibber-hk: Or 0.5.3 for the WithTag function if you choose to add that to IC.C

homothetyhk:
I've updated #21 for the unresolved conversations here.

ItemChanger.Silksong/Modules/AnonymousFleasModule.cs
flibber-hk: image

PR #55, opened 2026-02-15
Function to add feather effect

flibber-hk:
I think the API I've presented is a decent enough way to expose the feather effect, but maybe there's something better (or maybe it shouldn't even be exposed?). At the very least I think the feather effect belongs in IC even if it doesn't get used.

Obviously the numbers were taken from datamining (but I don't want to have to load from localpoolprefabs_assets_areagreymoor.bundle just to figure out the numbers at runtime)


flibber-hk:
See also https://discord.com/channels/879125729936298015/910615971222401084/1439532181302022225

homothetyhk:
I think the API looks good as is, though we can always revisit if needed. Thanks!

PR #59, opened 2026-02-17
Base game sprites

flibber-hk:
Result of the test:

image

UABEANext is recommended for getting the sprite names, and AssetHelperMenu[Dump asset names] to check that there is indeed only one asset inside the bundle (although I believe that is the case for all the base game sprite atlas bundles). 

This could be done with AssetHelper directly (create a managed asset, etc) but the sprite bundles have no dependencies so it loosens the timing requirements to do it like this.

### DIFF
--- a/ItemChanger.Silksong/ItemChangerPlugin.cs
+++ b/ItemChanger.Silksong/ItemChangerPlugin.cs
@@ -1,5 +1,6 @@
 using BepInEx;
 using ItemChanger.Silksong.Containers;
+using ItemChanger.Silksong.Serialization;
 
 namespace ItemChanger.Silksong
 {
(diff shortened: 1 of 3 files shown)


flibber-hk:
Alternatively base game sprites could simply be included as embedded pngs, personally I think this is easier though (ignoring the time I spent implementing AtlasSprite :zote:)

Also alternatively, most of these sprites can probably be reached using normal methods, e.g. the ventrica one as a descendant of the GameMap object, and the inventory ones from the inventory object.

flibber-hk:
> I'm afraid of asset bundles because I don't understand them very well

If it helps, I’ll write some notes in the IC discord channel later today that should be helpful for sprite stuff

PR #61, opened 2026-02-19
Add API for FSM edits

### LINKED ISSUE #53: Implement FSM edit API
Like in HK ItemChanger, we want to have a centralized API to handle FSM edits so that we only need to hook PlayMakerFSM.OnEnable once and also to save a little boilerplate on the edits themselves.

This API would be something like this:

```csharp
public static class FsmEdits
{
  public static void Add(FsmId id, Action<PlayMakerFSM> edit) { ... }
  public static void Remove(FsmId id, Action<PlayMakerFSM> edit) { ... }
}
```

dpinela:
This adds:

- AddFsmEdit and RemoveFsmEdit methods on SilksongHost
- SilksongHost.Instance static property to make those accessible to clients
- FsmEditGroup class to make it less boilerplate-y and error-prone to dispose of a bunch of FSM edits at once

FSM edits match an exact (scene, object, fsm) tuple, unless a wildcard is used in place of the scene or object. More complex match strategies (like a object name prefix match) are expected to use a wildcard and implement the match condition in the delegate itself.

Fixes #53


dpinela:
Yes, I considered having more sophisticated match options (like object name prefix matching, which you mentioned) but I decided it's simpler to just let clients define the criteria they need instead of trying to anticipate every use case.

So yes in the case you mention you'd wildcard the object name and then check for the path inside the delegate.

PR #67, opened 2026-02-23
Make crest item popups have the right size and text

### LINKED ISSUE #64: Implement UIDef for crests
The existing UIDef for crests (only tested the Hunter Crest so far) is broken; it shows a large sprite of the obtained crest and the word "Collected", and never goes away - this appears to be because crest items don't define a PopupName. Both of these should be rectified, and there should be a "big" version for these (and other major items as well; maybe that part should be spun off?) though the small one will always be needed.

dpinela:
There's still no icon for the Cloakless Crest, but I don't even know what icon that should have.

For reference, this is what the Hunter crest popup looks like now:
hunter at wanderer

Fixes #64

### DIFF
--- a/ItemChanger.Silksong/Items/ItemChangerSavedItem.cs
+++ b/ItemChanger.Silksong/Items/ItemChangerSavedItem.cs
@@ -1,4 +1,5 @@
 ﻿using ItemChanger.Items;
+using ItemChanger.Serialization;
 using ItemChanger.Silksong.Serialization;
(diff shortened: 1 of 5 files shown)


dpinela:
The problem with both of these is that concatenating these ends up with "Crest X" instead of "Crest of X". I tried Tools/UI_MSG_TITLE_CREST but that gives me "Bound the Crest of X" which is also not what I wanted.

I agree that the solution you suggest for bellways would solve this problem as well.

PR #68, opened 2026-02-23
Non-Eva crest locations

### LINKED ISSUE #62: Implement non-Eva crest locations
This would encompass the bindable corpses at the Chapels of the Wanderer, Beast, Reaper, and Architect, and the Ruined Chapel, plus the Yarnaby cure sequence.

Leaving Eva out of this one since it's not immediately necessary to rando/swap crests and also the implementation will have to be entirely different anyway.

dpinela:
Some or all of the FSM extensions may be worth upstreaming to FsmUtil. Not sure whether that should block this.

Fixes #62


flibber-hk:
SFG is generally pretty good at merging and releasing updates to FsmUtil quickly, at least in my experience, so you could probably get updates included there if you wanted to make a pr

dpinela:
MustGetState could in practice be replaced with `GetState(...)!`, though raising the exception earlier is nice. The `Index*` and `*Matching` methods would definitely be good to upstream, nothing in FsmUtil offers that kind of functionality right now.

dpinela:
Opened https://github.com/silksong-modding/Silksong.FsmUtil/pull/20 for upstreaming the FSM extensions added here.

BadMagic100:
It's probably worthwhile to spot check one and see if the name should be changed

PR #78, opened 2026-03-04
Feature/issue 7 chest container

tobiaslory:
Implements chest container support for #7.

What’s in:
- `ChestContainer` is now actually implemented (instantiate + modify in place).
- New `ChestContainerBehaviour` handles opening, persistence, and spawning rewards.
- Chests now spawn one shiny per unobtained item for multi-item placements.
- Added chest asset requests in startup.
- Added `MultiItemChest_in_Tut_02` test in `ItemChangerTesting`.
- Included the small test-dispatcher fix so test profile setup is applied reliably.

Prefab/source behavior:
- Prefer native `Hang_06_bank` chest paths.
- Keep `Tut_01/Bone Chest` as fallback so we still have a known working option if bank paths fail.

Gameplay behavior notes:
- Chest opens from attack-like collider hits (not just walking into it).
- Vanilla reward object under chest `Item` is neutralized to avoid duplicate reward paths.
- Open state visuals now prioritize native chest FSM behavior, with fallback visuals only when needed.

Validation:
- Build passes.
- In-game chest test passes (correct chest source resolves, chest opens, expected shiny count spawns).

### DIFF
--- a/ItemChanger.Silksong/Containers/ChestContainerBehaviour.cs
+++ b/ItemChanger.Silksong/Containers/ChestContainerBehaviour.cs
@@ -1,5 +1,4 @@
 using System.Collections;
-using System.Reflection;
 using HutongGames.PlayMaker;
 using ItemChanger.Containers;
 using ItemChanger.Enums;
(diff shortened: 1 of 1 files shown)


flibber-hk:
Out of curiosity, is there a reason you wrote a custom component rather than modifying the chest fsm? My assumption would be that the fsm edit would be a lot easier

PR #82, opened 2026-03-10
Add Grounded Sprint novelty item

DatUub:
Adds a **Grounded Sprint** novelty item that gives sprint-only functionality (no dash) when on the ground, reading `DASH_SPEED` from `HeroController`.

## Changes
- **`GroundedSprintModule.cs`** (`CustomSkills/`): Listens for sprint input (`inputActions.Dash.IsPressed`) and boosts ground velocity to `DASH_SPEED` via `HeroController.Move` postfix.
- **`Novelties.cs`**: Registers `Grounded_Sprint` as a `CustomSkillItem` with `ItemChainTag` (successor: `Swift_Step`).
- **`ItemNames.cs`**: Added `Grounded_Sprint` constant.
- **`GroundedSprintTest.cs`**: Test placing Grounded Sprint twice; first pickup gives ground sprint, second auto-advances to Swift Step via chain.

## Progression
Grounded Sprint -> Swift Step via `ItemChainTag`. First pickup gives ground-only sprint (velocity boost when holding sprint key on ground). Second pickup advances to full Swift Step with dash.

### DIFF
--- a/ItemChanger.Silksong/Modules/CustomSkills/GroundedSprintModule.cs
+++ b/ItemChanger.Silksong/Modules/CustomSkills/GroundedSprintModule.cs
@@ -59,7 +59,7 @@ public class GroundedSprintModule : CustomSkillModule
         if (!InputHandler.Instance.inputActions.Dash.IsPressed) return;
 
         Rigidbody2D rb = __instance.GetComponent<Rigidbody2D>();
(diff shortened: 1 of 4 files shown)


DatUub:
so should gorunded sprint also give the dash as well? but only when grounded?

Issue #97, opened 2026-03-15
Implement tablet container


homothetyhk:
This container would be used to modify or instantiate lore tablets (compare Hollow Knight's [Tablet Container](https://github.com/homothetyhk/HollowKnight.ItemChanger/blob/master/ItemChanger/Containers/TabletContainer.cs) and [TabletUtility](https://github.com/homothetyhk/HollowKnight.ItemChanger/blob/master/ItemChanger/Util/TabletUtility.cs)).

Also, the container should provide a static method to create a tablet with a given `Func<string>` message provider. The value of the message should be refetched using the delegate whenever the tablet is read.

flibber-hk:
I'll do the container implementation for existing lore tablets - this will be blocked by #115 and will probably need some merge conflict resolution with #101 

(I probably won't finish off the existing lore tablets in BaseLocationList though)

homothetyhk:
With #192 (including #159), the container is essentially done, but tablet ObjectLocations do not currently support replacement, due to lack of DestroyOnContainerReplaceTags. Ideally, corrections should be measured in the same PR.

PR #101, opened 2026-03-16
Eva location

### LINKED ISSUE #63: Implement Eva location(s)
Obvious companion to #62 

This should cover the scope of the vanilla hunter upgrades, vesticrests, and sylphsong. Costs should be handled transparently by ItemChanger rather than using the vanilla locket costs. This could plausibly be implemented similarly to Grubfather and Seer in HKIC (which would warrant the implementation of a similar CostChestLocation)

dpinela:
Eva functions as a shop location, able to accept any number of items and displaying a list of obtainable items at a nearby lore tablet, similarly to Seer or Grubfather in Hollow Knight. She may also be configured to retain some or all of her vanilla items by setting DefaultItems appropriately on the placement.

A new cost type is added to represent tool slot costs, primarily for use with this location; to avoid complications with rando logic, these costs count tool slots unlocked for the Hunter Crest, unlike vanilla costs.

Closes #63


dpinela:
I've spun off the tablet-instantion method over to the LoreTabletContainer class according to https://github.com/homothetyhk/ItemChanger.Silksong/issues/97#issue-4078805991. Leaving that class as a stub otherwise for now. This should now be ready to review/merge.

dpinela:
Also this introduces a new asset cache class for ManagedAssetLists, instead of special-casing the Weave_08 lore tablet as was suggested earlier; I think that makes more sense anyway.

flibber-hk:
> Also this introduces a new asset cache class for ManagedAssetLists, instead of special-casing the Weave_08 lore tablet as was suggested earlier; I think that makes more sense anyway.

FWIW my rationale for saying that game object lists would go in the game object cache was that the game object cache would be the one who knows that to get the tablet group you have to get the list corresponding to the tablet and select the one you need, and the client (in this case the tablet container) just calls the AssetCache.InstantiateAsset method without needing to know the details of whether it’s loaded as a list rather than as a GameObject; that said I haven’t thought about it enough to confidently opine which way is actually better

dpinela:
I thought for that to work, I'd need to come up with some syntax for denoting that these methods need to look for a specific subobject, but now that I think of it that's not the case and I had another idea.

dpinela:
I implemented an alternative solution that works like you described. Because of the fact that the criterion for selecting an object out of a ManagedAssetList is an arbitrary predicate, these assets can't be specified fully in JSON, so the solution is a little janky. Overall it looks better than before but not by all that much.

flibber-hk:
Does/should this PR include a mechanism whereby rando can get notified when they preview the location, for helper log purposes? I can’t remember how this is supposed to work, but IIRC the ILocalHintLocation interface needs to be involved somewhere

dpinela:
IC already has the ability to mark placements as previewed, I just need to use it.

dpinela:
I also noticed that the implementation doesn't honour DisableCostPreviewTag, I should add that as well.

PR #104, opened 2026-03-22
Ic object cache

### LINKED ISSUE #76: Set up centralized object cache
In particular, some way of requesting and loading/unloading managed assets that don't necessarily have a proper home. (The flea asset should be moved from the container to here, and optionally also the flea container presentations depending on maintainer/implementer taste). (Note - IIRC the flea asset is used both for the sprite and to count the number of fleas rescued so it probably can't be removed).

It could be implemented as a default module (or a collection of default modules based on a generic type parameter), or a repository (like in HK.IC) which is loaded/unloaded directly by the host.

flibber-hk:
Closes #76 

I think it's impossible to do this without some cursed casting, but I think this is a good way of doing things in that there isn't a huge amount of work involved in adding new assets.

I was considering doing the big item popup but I didn't realise it would involve a lot more infrastructure work so I left it; I put the asset in the json because I wanted to have an example, but I wouldn't be surprised if other non-scene game objects get loaded first (in which case you may or may not want to remove that asset until it gets needed).


homothetyhk:
Hooking `GameManager.ReturnToMainMenu` doesn't guarantee `OnLeaveGame` is invoked before unload, when unloading due to Benchwarp (`gm.LoadGameFromUI`). So I think I'll keep the fix I did when resolving the merge conflicts.

flibber-hk:
Based on LTSmash's comment in #itemchanger in the discord, I have commented out the other chests because I think having a sequence of fallbacks isn't really necessary.

PR #115, opened 2026-03-27
Default big UI def

flibber-hk:
This is a partial resolution to the big uidefs issue which I can't name else github may try to close the issue. In particular:

* It is possible to create an arbitrary big UI def in the base game style - the CustomDefaultBigUIDef and DefaultBigUIDefData classes can be extended if there's any missing functionality
* BigUIDefs for some - but not all - items needing them have been added. Notably the crests haven't been done, see below.

Notes:
* I'm using the FSM prompt because it's easier to make custom. The non-FSM prompts, which inherit from `UIMsgBase<>`, are controlled by an animator (I believe) - those are probably harder to modify at runtime, but I think the speed property of the animator could be used to speed them up. (I think there would be some more hooks involved). Whether they're needed depends on how people choose to do crests - in principle it should be easy to spawn them by doing ToolCrestUIMsg.Spawn with the ToolCrest as an argument (and the correct prefab loaded from prompts_assets_all.bundle), but I decided I'm leaving that out of scope.
* I haven't done the UIDef for Taunt because I didn't want to write text for it :zote: the taunt test works though
* I went for the alt sprites where possible for spells because I like them more, unfortunately this only does 4 of them


flibber-hk:
Note to self: big ui defs shouldn’t be supported for instant shinies :zotesip:

flibber-hk:
I wonder if we should make it so that if you take damage, subsequent big UI defs don't display. We could reasonably do this by rebranding CascadingUIDef as ControlRelinquishedUIDef, and having it automatically fall through to the fallback if control is currently not relinquished.

flibber-hk:
I think I'm gonna close this PR, so I don't need to ferry changes between this branch and #159

PR #116, opened 2026-03-29
Add default ObstacleSuppressionModule

### LINKED ISSUE #96: Module to open crest chapels permanently
If any item is placed within a chapel, that chapel must stay open at least long enough for everything in it to be collected. Given that connections may add more locations within chapels, it's safest to just keep them open permanently.

Things I'm not sure about:

- Should the module let you choose which chapels to apply to, or open them all unconditionally?
- Should it be a default module, or left for rando (and other client mods) to add and configure?

dpinela:
Though only the crest chapels are included for now, the module is general enough to include other obstacles that may be suppressed like the Far Fields lava challenge.

Resolves #96

### DIFF
--- a/ItemChanger.Silksong/Modules/ObstacleSuppressionModule.cs
+++ b/ItemChanger.Silksong/Modules/ObstacleSuppressionModule.cs
@@ -1,0 +1,99 @@
+using ItemChanger.Modules;
+using HutongGames.PlayMaker;
+using Benchwarp.Data;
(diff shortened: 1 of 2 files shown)


homothetyhk:
I've modified the implementation to use Benchwarp's data as a base, and to more easily allow injecting/suppressing obstacles. The API is not the best, but hopefully it will get the job done. We can revisit later if it becomes too cumbersome.

PR #121, opened 2026-03-30
Autoequip silk skills granted at crest corpse locations

### LINKED ISSUE #120: Vanilla silk skills don't autoequip after collecting a CorpseCrestLocation
How to reproduce:

- Place any ~crest (that can take silk skills)~ item at a corpse location, like the Reaper Crest location
- Collect that ~crest~ item
- ~Sit at a bench and equip the new crest~
- Collect any vanilla silk skill (I've specifically tested Silkspear and Thread Storm; others likely work similarly)

At this point, the skill should be auto-equipped and ready for use, but it isn't. Saving and quitting fixes this state, but benchwarping, interestingly, does _not_.

This issue appears to be specific to CrestCorpseLocations. ~crests obtained from shinies or from Eva (using the implementation in #101) work as expected in this situation.~

tobiaslory:
closes #120


flibber-hk:
I’m pretty sure you’ve misunderstood the issue driving #120? The problem is that collecting anything at a crest corpse location breaks something that means future silk skills don’t work properly, the silk skill doesn’t have to be at the crest corpse location

(at any rate, I can’t see how the code you’ve provided handles the case described in that issue)

dpinela:
Correct. Not only that, but the behaviour this PR introduces doesn't make sense even if we did want to autoequip silk skills - if we did that, we'd want it to apply to skills acquired anywhere, not just at crest corpses.

tobiaslory:
I've updated based on your feedback, I couldn't identify any issue specific to the CrestCorpseLocation, so I made the auto equip behaviour global

flibber-hk:
> if we did that, we'd want it to apply to skills acquired anywhere

Do we actually want to do this? Does it auto-equip silk skills in the base game? (if yes, what if Architect crest is equipped?) I feel that we probably would prefer not to for logic reasons but I haven't really thought through the ramifications of either option

dpinela:
The base game does auto-equip silk skills (and crests); however in a rando it would indeed be problematic to do so, both for logic reasons and because it you could get them anywhere so it would be rather disruptive.

homothetyhk:
Autoequip is nice to have available, but for the reasons already mentioned, it should not be a default--the IC consumer should have to intentionally opt-in to it. In HK, the way we did it was through a [special tag](https://github.com/homothetyhk/HollowKnight.ItemChanger/blob/master/ItemChanger/Tags/EquipCharmOnGiveTag.cs) which changed the behavior of CharmItem to autoequip. I think a similar item tag would make sense for Silksong (ideally, it would be nice to have autoequip for silk skills, tools, and crests, though the situation is more complicated for tools).

I'll close this for now, but I would accept a PR for the tag.

PR #153, opened 2026-04-06
Implement Weaver Burial Spire locations

### LINKED ISSUE #146: Implement Weaver Burial Spine locations
These locations are controlled by FSMs that appear to be similar if not identical to each other, as at least one of them references all the possible abilities that you could get (similar to crest corpse FSMs).

thomaslunn:
Adds locations for each of the Weaver Burial Spires (except First Sinner). These are:

- Silkspear
- Swift Step
- Thread Storm
- Sharpdart
- Cling Grip
- Clawline
- Silk Soar

The initial animation sequence is preserved, but the fade-to-black is shortened slightly. There's capability here to maintain the ability pop-up when a silk skill or ability is picked up at a shrine - maybe this is better handled by the item rather than the location.

Resolves #146.


flibber-hk:
I wonder if this should be done as a container? That way it can e.g. be replaced by a flea when there's a flea there

flibber-hk:
> There's capability here to maintain the ability pop-up when a silk skill or ability is picked up at a shrine - maybe this is better handled by the item rather than the location.

See #115 for an initial version of this (handled by the item rather than location) - I don't know enough about these locations to know what (if anything) needs to be done to make them support big UI defs properly

thomaslunn:
> I wonder if this should be done as a container? That way it can e.g. be replaced by a flea when there's a flea there

That sounds fun. Would we also want Weavers appearing in other locations?

flibber-hk:
> Would we also want Weavers appearing in other locations?

Hmm, maybe it depends on how they look? I would have assumed no because it doesn't happen in HK but it could maybe be done

dpinela:
If Weavers wouldn't appear in other locations, it doesn't make sense to make this a container, unless I misunderstood something?

thomaslunn:
I've updated to use a container with `SupportsInstantiate = false` - so Weaver shrines can have their locations replaced e.g. by fleas, `Shiny` items will continue to use the Weaver shrine, and corpses won't appear around the world.

thomaslunn:
Note this PR will need an update once ItemChanger.Core 0.9.2 is merged

homothetyhk:
I think the fsm edit needs to make a few changes to some of the later states (End, Set Finished). The main idea is that we should get the same overall effect when giving items whether the container is replaced or not. Right now the spire has a few side effects (sets respawn, gives silk). Probably these should all be removed (and maybe at a later date we can decide whether to add the silk effects to the original items).

Also, with the new IC.Core update, you should add an `info.ApplyTo` to the start of `ModifyContainerInPlace`.

homothetyhk:
What are the practical effects of the Silk Soar escape sequence? It seems to me just that the platform is destroyed after a successful Silk Soar or about 5 seconds.

Then, a few notes:
- The retrigger conditions for the sequence are not modified. So, the sequence will retrigger if the player has Silk Soar and has respawn in Abyss_08, and the platform will stay destroyed if the player has Silk Soar and respawn outside Abyss_08.
- This might make it difficult to collect respawning items for various choices of containers, so we probably need to enforce that the platform doesn't get destroyed.
- Triggering the sequence by visit state isn't ideal since it means that the container is modified and any of its items have big popup / dialog, then the platform can break before the item animations finish.

I'm not sure there's a way to handle it that doesn't make it very painful for something along the lines of a chest container. With that in mind, I think the best option is either remove the sequence entirely, or force the container to be WeaverCorpse (and prevent the sequence complete check when the player has Silk Soar). Either choice is fine by me.

ItemChanger.Silksong/Tags/RaiseFsmEventOnGiveTag.cs
```diff
@@ -0,0 +1,99 @@
...
+            placement.OnVisited -= OnVisited;
+        }
+
+        ItemChangerHost.Singleton.GameEvents.RemoveSceneEdit(SceneName, FindFsm);
+        _fsm = null;
+    }
+
+    public void OnReplace(Scene scene, GameObject newContainer)
+    {
+        _wasReplaced = true;
```
homothetyhk, line 78: `_wasReplaced` should probably be reset somewhere, in case some change at runtime causes the placement to select a new container when reloading the room. It might be safe to do this in the scene change edit (location tags before the parent DoLoad), or you could do it in BeforeNextSceneLoaded to be extra safe.

homothetyhk:
I finished my review, so I can merge once the few things I mentioned are good. Thanks for the PR!

thomaslunn:
I've gone ahead and made all the requested changes, so as long as you're happy this is ready to merge.

Good point on the persistent items handling, I hadn't fully realised how the fsm was working on room reload. I've gone ahead and removed the escape sequence entirely for now.

PR #156, opened 2026-04-07
Add ObjectLocations for world Mask Shards and Spool Fragments (#110)

### LINKED ISSUE #110: Implement locations for interactable Mask Shards and Spool Fragments
This concerns shards/fragments which are acquired as floating interactables (i.e. excluding shop items, quest rewards, etc, which will require different implementations). Interactables spawned on event (e.g. defeating a boss, etc) may require additional work to implement.

These locations are expected to be implemented as ObjectLocations with no original container, since a container implementation for mask shards/spool fragments is not planned. Hence, the location will act to replace the interactable with a shiny item or other suitable container as selected by the placement.

kickafetus:
Implements all world-pickup Mask Shard (14) and Spool Fragment (13) locations as ObjectLocations, excluding shop and quest-reward items which are handled separately.

Each location uses FloatShiny — a ShinyControlTag combining ShinyType.Instant (contact pickup, same mechanic as mossberry drops) with ShinyFling.FloatInPlace (kinematic rigidbody, no gravity) — so the replacement shiny hovers exactly where the original interactable was without altering pickup difficulty.

ShinyContainer.ModifyContainerInPlace gains ShinyDetachScheduler for FloatInPlace shinies: a DontDestroyOnLoad MonoBehaviour that unparents the shiny from any container hierarchy in LateUpdate. This handles Heart Piece objects parented to breakable containers (e.g. Library_05 jar, Bilewater cocoon) that would otherwise destroy the shiny when broken. Start()-based detach was unreliable because ApplyTargetContext can call SetActive(false), which cancels pending Start() calls.

Closes #110

### DIFF
--- a/ItemChanger.Silksong/Containers/ShinyContainer.cs
+++ b/ItemChanger.Silksong/Containers/ShinyContainer.cs
@@ -4,7 +4,9 @@ using ItemChanger.Extensions;
 using ItemChanger.Silksong.Extensions;
 using ItemChanger.Silksong.Tags;
 using Silksong.UnityHelper.Extensions;
+using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
 using UnityEngine;
 
 namespace ItemChanger.Silksong.Containers;
@@ -158,6 +160,28 @@ public class ShinyContainer : Container
             shiny.fling = false;
             rb.bodyType = RigidbodyType2D.Kinematic;
             shiny.pickupAnim = CollectableItemPickup.PickupAnimations.Stand;
+            // FloatInPlace shinies are already static — the prefab's waitForStoppedMoving
+            // behavior (designed for flung shinies that need to land) would temporarily
+            // deactivate interactEvents and only re-enable it once the rigidbody velocity
+            // drops below 0.1. Since the body is kinematic (velocity always zero), this
+            // would re-enable after canPickupDelay — but we want immediate interactability.
+            // Disable waitForStoppedMoving and zero canPickupDelay so interactEvents
+            // is never suppressed and the shiny is pickable the moment it appears.
+            Type cipType = typeof(CollectableItemPickup);
+            cipType.GetField("waitForStoppedMoving", BindingFlags.NonPublic | BindingFlags.Instance)
+                   ?.SetValue(shiny, false);
+            cipType.GetField("canPickupDelay", BindingFlags.NonPublic | BindingFlags.Instance)
+                   ?.SetValue(shiny, 0f);
+
+            // Some Heart Piece / Silk Spool objects are parented to (or start inactive inside)
+            // a breakable container (e.g. Library_05 jar). ObjectLocation parents our
+            // replacement shiny to that same container and mirrors the Heart Piece's active
+            // state. If the Heart Piece was inactive, Unity cancels the shiny's Start()
+            // call when SetActive(false) is applied — so a Start()-based detach never runs.
+            // ShinyDetachScheduler uses a DontDestroyOnLoad MonoBehaviour whose LateUpdate
+            // fires reliably after sceneLoaded (when IC edits run) regardless of active state,
+            // detaching the shiny from the container and ensuring it is active.
+            ShinyDetachScheduler.Schedule(shiny.gameObject);
         }
         else
         {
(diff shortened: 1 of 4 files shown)


flibber-hk:
Apologies if I'm missing something, but why is your change needed to ShinyContainer.cs? They all seem to be root game objects in your BaseLocationList files, so deparenting shouldn't be needed

flibber-hk:
Um, also is it the behaviour we actually want that shinies should float? I feel it would make more sense if the shinies were normal shinies on the ground where it doesn't impact logic

kickafetus:
How would it floating affect any real logic? I sort of took the idea of them staying in place to just match where they already are.

And the mask shard that is in scene library_05 would keep disappearing each time I would destroy the little container that it is in. I guess I could go the route of not having the container hold the mask/item anymore and just have something there, but I liked the idea of it being in there anyway like the game has it, and to keep it as close to the game as possible.

flibber-hk:
> How would it floating affect any real logic? I sort of took the idea of them staying in place to just match where they already are

Sorry, when I said logic I meant the randomizer logic (which items can allow you to get each location). As I mentioned on the original issue #110, we need to make the mask shard in Mount Fay float because if that one's on the ground, the logic changes - you no longer need double jump to get it if it's on the ground. My opinion is that outside of cases like this, shinies should fall to match vanilla shiny behaviour; if you think it actually looks better this way then I’ll defer to you/Homothety

> the mask shard that is in scene library_05 would keep disappearing

Hmm, it seems this one probably requires a custom implementation, because the shiny probably should be made to appear when the container breaks (rather than simply being there at the start). Are there any other locations in this update that have the same property? (This would require more coding work so depending on how much you want to do, you could either have a go at getting this to work, or leave comments in the BaseLocationList files indicating which ones need additional work)

kickafetus:
From what I can remember, all of the other shards seem to still be perfectly reachable and thus why I sort of just defaulted to wanting it to be closer to how the game has it. So unless I am mistaken, I don't think they should affect the randomizer logic?
Also I know that there are a few spools that would be precarious, once allowed to fall lol (or at least I would think, like the top of high halls and the one above water in weavenest atla)

For the spawning on break, that's a fair point. I do kind of have it being a half jank solution.
I don't believe there are any other items that are part of the spools or masks that have this characteristic.
I can still try and tackle it in the next couple of days while I am also working on the wishboards.

flibber-hk:
> So unless I am mistaken, I don't think they should affect the randomizer logic

Oh, my original message probably was poorly written. Obviously floating shinies don’t affect logic. What I meant was, if normal (fallen) shinies also don’t affect logic, then they would make more sense

homothetyhk:
I think the main reason we prefer non-contact shinies is that they can deliver UIDefs that are supposed to be used while control is removed (big item popups, lore dialogue, and so on). That probably is reason enough to use normal shinies wherever it is safe to do so.

In HK there were a few locations that used those types of UIDefs while the player had control, but there is some risk involved with potential of entering glitched states while that happens, so some investigation would be needed before making a general rule for Silksong.

flibber-hk:
FWIW I'm pretty sure that detaching the shiny from the parent is incorrect in general - if the original object was inactive, then the new shiny should remain inactive (with custom logic to activate it when the vanilla object should be activated).

For the locations you've implemented, I'd guess they won't work properly for containers other than shinies (see e.g. if you place a flea there). Depending on how much work you want to do, you probably ought to either make them force shiny (by setting ForceDefaultContainer to true) or set the elevation (by setting the Correction to be a Vector2 whose y coordinate is the y coordinate of the object in unity explorer minus the y coordinate of hornet when she's there).

kickafetus:
>In HK there were a few locations that used those types of UIDefs while the player had control, but there is some risk involved with potential of entering glitched states while that happens, so some investigation would be needed before making a general rule for Silksong.

This makes total sense, and I'm perfectly fine with resolving this, in one way or another for the positions where they don't require to be floating.
I have actually seen it happen when I tested the Spool in Weavenest, I would just continually fall and then would be teleported back to ground when the animation stopped. My apologies for not being able to pay more attention there and put two and two together lol

>FWIW I'm pretty sure that detaching the shiny from the parent is incorrect in general - if the original object was inactive, then the new shiny should remain inactive (with custom logic to activate it when the vanilla object should be activated).

Heard loud and clear.
Also absolutely happy to look into reworking and testing with other container types. Just off the top of my head, I'll probably end up seeing about forcing shiny, but if I come across anything else or have any larger questions, I'll make sure to ask.

When I do get back to library_05, I was thinking today, that it might just be easiest to have the container be broken on room entry and then just having the shiny on the ground already. No additional call for release needed then. Part of me still wants to see about messing with the object and potentially work out an elegant solution to fit within the game and also mayyyyyybe show what might be in the case before breaking (idk about that one, actually just well wishing without knowing the level of effort lol).

I will go ahead and cancel this PR, and look to make some of these changes and then bring them back up in the issue thread before just submitting another PR.

Thanks for all of your explanations and suggestions here.

PR #159, opened 2026-04-09
Tablets

### LINKED ISSUE #157: Lore UI Def
I am claiming this

flibber-hk:
Included: Tablet items and locations

Some TODOs have been noted in the source

~~Blocked by~~ Encompasses #115 

Closes #157 and #97


flibber-hk:
Also TODO would be journal/materium locations, I haven't done these but they should be doable with the same container

flibber-hk:
Converting to draft because I'll work on a couple of the TODOs before converting back to non-draft

homothetyhk:
From the other thread:
> I wonder if we should make it so that if you take damage, subsequent big UI defs don't display. We could reasonably do this by rebranding CascadingUIDef as ControlRelinquishedUIDef, and having it automatically fall through to the fallback if control is currently not relinquished.

This probably doesn't change the potential for glitches too much, since someone can still carry the first big UI def around and get into whatever kind of trouble with it. But it does sound like nice QoL, since I think it is unpleasant to have to mash through multiple popups while being attacked.

flibber-hk:
Converting to draft because I found another bug :zotesip:

PR #161, opened 2026-04-09
Implement currency items

dplochcoder:
Shell shards don't appear to support 'fling type = straight-up' very well, possibly because they have larger non-circular collider boxes that can clip into the ground on spawn.  Haven't debugged in depth, assuming it's low priority.  Everything else works as expected.

### DIFF
--- a/ItemChanger.Silksong/ItemChanger.Silksong.csproj
+++ b/ItemChanger.Silksong/ItemChanger.Silksong.csproj
@@ -97,18 +97,6 @@
     <PackageReference Include="Silksong.Benchwarp" Version="1.1.3" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(Configuration)' == 'Debug'">
-    <Reference Include="MMHOOK_Assembly-CSharp">
-      <HintPath>$(SilksongFolder)/BepInEx/plugins/MMHOOK/Managed/MMHOOK_Assembly-CSharp.dll</HintPath>
-    </Reference>
-    <Reference Include="MMHOOK_PlayMaker">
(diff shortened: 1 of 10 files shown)


homothetyhk:
What's the reasoning behind SilksongFolder to SilksongPluginsFolder? Is it just to allow some future flexibility for the csproj to refer to the Managed folder independently from local install / copy mod dir? I'm confused because there's still an `Exists('$(SilksongFolder)')` in the PR, but the path itself is never used.

More to the point, I think all of the MMHOOK references can actually be removed from the csproj, now that we use MonoDetour. In that case, the only use for the path would be to specify the destination of CopyMod.

dplochcoder:
> What's the reasoning behind SilksongFolder to SilksongPluginsFolder? Is it just to allow some future flexibility for the csproj to refer to the Managed folder independently from local install / copy mod dir? I'm confused because there's still an `Exists('$(SilksongFolder)')` in the PR, but the path itself is never used.

The two variables have existed in the silksong modding template for some time now, it's not a future thing.  The distinction is necessary if you want to (1) use r2modman or otherwise install plugins separately from the Silksong executable *and* (2) reference the local Silksong libraries during development.

ModMenu has [this](https://github.com/silksong-modding/Silksong.ModMenu/blob/451ea90e9f813c357566362419d0aac81a69d6b9/Silksong.ModMenu/Silksong.ModMenu.csproj#L51-L65) drag-and-drop block that makes it possible to do this easily, but it doesn't work when the csproj otherwise assumes 'SilksongPluginsFolder == $(SilksongFolder)/BepInEx/plugins'.

> 
> More to the point, I think all of the MMHOOK references can actually be removed from the csproj, now that we use MonoDetour. In that case, the only use for the path would be to specify the destination of CopyMod.

Done.

PR #162, opened 2026-04-10
Feature/issue 109 wishwall location

### LINKED ISSUE #109: Implement WishwallLocation
We need a location type for quests that give a reward directly from the quest board (wishwall). For example, this would include most Hunt wishes and Grand Hunt wishes. I think this is likely best done as an `AutoLocation`, with one location/placement per quest (hence potentially several placements at the same quest board). A few notes:
- The implementation must support respawning items. On interacting with the quest board, it should give all refreshed items for all placements associated with it, and subsequently allow normal use.
- The implementation should support previews (though this can be split off from the initial implementation if it is too difficult). The sprite of the reward item in the quest listing should be replaced (ideally with text from concatenating `UIDef.GetPreviewName` for the items in the placement and truncating to a safe length, but some fiddling may be needed to find an aesthetically pleasing way to achieve this).

Beyond scope: 
- Quests completed by talking to NPCs, as they may require more specific hooks and fsm edits to control the NPC state.
- Editing unlock conditions and completion requirements for quests. This will likely be handled through a separate module.

kickafetus:
Closes #109

### DIFF
--- a/ItemChanger.Silksong/Locations/WishwallLocation.cs
+++ b/ItemChanger.Silksong/Locations/WishwallLocation.cs
@@ -15,25 +15,11 @@ namespace ItemChanger.Silksong.Locations;
 /// instead of the quest's vanilla reward.
 /// </summary>
 /// <remarks>
-/// <para>
 /// <see cref="QuestName"/> must equal the Unity Object name of the <see cref="FullQuestBase"/>
-/// ScriptableObject — this is the <c>QuestCompletionDataName</c> field in
-/// <c>quests.json</c>, and can be referenced via the generated
-/// <see cref="RawData.Quests"/> constants.
-/// </para>
-/// <para>
-/// Reward-icon and description preview (hooking <see cref="FullQuestBase.RewardIcon"/>,
-/// <see cref="FullQuestBase.RewardIconType"/>, <see cref="FullQuestBase.GetDescription"/>,
-/// and <see cref="QuestItemDescription"/>) is handled centrally by
-/// <see cref="WishwallPreviewModule"/>, which this location registers with on load.
-/// </para>
-/// <para>
-/// The hook intercepts <see cref="FullQuestBase.TryEndQuest"/> globally and wraps
-/// the <c>afterPrompt</c> callback to inject <c>GiveAll()</c>. This approach works
-/// for all completion paths: both the standard <c>ProcessQueuedCompletions</c> route
-/// and FSM-driven donation quests that call <c>TryEndQuest</c> directly with an
-/// <c>afterPrompt</c> that does not invoke <c>rewardItem.Get()</c>.
-/// </para>
+/// ScriptableObject — the <c>QuestCompletionDataName</c> field in <c>quests.json</c>.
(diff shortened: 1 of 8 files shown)


flibber-hk:
Just wondering, should these quest locations be identified as such, or should they be identified by the item? e.g. `Wishwall__Beastfly_Hunt` could instead be `Mask_Shard__Savage_Beastfly`

kickafetus:
lol...
I was too worried about getting the wishwalls working how we wanted, I completely forgot to worry about the actual rewards and items attached to them

I will correct that.
Thank you flibber

kickafetus:
Do you want me to still have the rest of the quests listed out how I do?
I just renamed the 10 in quests.cs and removed the duplicates from location names, that directly correlate to what is requested in #109: Quests that have rewards when interacting directly back with the Wishwall

[Commit for the quest renames](https://github.com/homothetyhk/ItemChanger.Silksong/commit/c6448e17aeb995ca3326a092f477d88e68e45a7f)

Because of my inexperience I am not sure if the additional listing of the quests is useful or not, so I just wanted to check and see before I axe them.

homothetyhk:
Here's a small bug, related to quests with prefab rewards (e.g. mask shard reward).
- If I call `QuestUtil.SetReadyToComplete(Quests.Ant_Trapper)` and then interact with the Bellhart wishwall (`PlayerData.instance.spinnerDefeated = true`, with an IC placement at `Mask_Shard-Gurr`), then I receive the IC items, followed by the vanilla mask shard reward.
- On the other hand, if I give a Grass Doll at start and only set the quest to seen and accepted, then the wishwall only gives the IC items.

To summarize the issue, collecting the vanilla Grass Doll with the quest active puts the quest into its ready to complete state, and when it is in this state, `QuestBoardInteractable.ProcessQueuedCompletions` produces the mask shard as a prespawned item, separately from the reward item.

To fix it, I think we would want to either: patch the quest's reward item before the quest board awakes (which is when it first checks for prefabs), or ignore the `afterPrompt` passed into `TryEndQuest` and call the `ExecuteDelayed(1f, ProcessQueuedCompletions)` ourselves. Let me know what you think.

kickafetus:
Hello! Apologies for the delayed response!
Needed to get reacquainted with what I did in here lol

Initially I would think we would want to just catch it before the awake properly maps `ISavedItemPreSpawn` into `preSpawnedItems` and map our `rewardItem` then just let it go back to running normally with the orig? So I am all for your first idea.
`AccessTools.Method(typeof(QuestBoardInteractable), "Awake"),
            (Action<QuestBoardInteractable> orig, QuestBoardInteractable self) =>
            {
            *map the item stuffs*
            }`

Would that work? My only question here would be if we let it run normally after our mapping, would it just overwrite what we wrote before or would it skip that portion because it's already mapped?

homothetyhk:
I think hooking awake should be fine--once the reward item is overriden on the quest, everything else looks ok since the modded reward item won't have ISavedItemPreSpawn (I haven't tested though). I guess this makes some of the code in InterceptTryEndQuest redundant, but it looks like it should still run with the quest already patched without issue.

PR #165, opened 2026-04-10
Test plugin improvements

dplochcoder:
The `DynamicChoiceModel` class adds a lot of complexity that is not necessary now that `ListChoiceModel.UpdateValues()` exists (it's possible it didn't when this was first written).  Removing this class makes the menu logistics much easier to understand.

### DIFF
--- a/ItemChangerTesting/ItemChangerTestingPlugin.cs
+++ b/ItemChangerTesting/ItemChangerTestingPlugin.cs
@@ -1,5 +1,6 @@
 ﻿using BepInEx;
 using BepInEx.Configuration;
+using HutongGames.PlayMaker.Actions;
 using ItemChanger;
 using ItemChanger.Events;
 using ItemChanger.Silksong;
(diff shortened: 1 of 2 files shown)


ItemChangerTesting/ItemChangerTestingPlugin.cs
```diff
@@ -38,18 +40,27 @@ public AbstractMenuScreen BuildCustomMenu()
...
+            ListChoiceModel<Test> model = new([.. Test.TestGroups[cfgTestFolder.Value]])
+            {
+                DisplayFn = (_, t) => t.GetMetadata().MenuName
+            };
             DynamicDescriptionChoiceElement<Test> testSelector = new("Test", model, "The test to launch.", t => t.GetMetadata().MenuDescription);
 
             TextButton run = new("Erase save slot and launch test.");
             run.OnSubmit += Run;
+
+            TextButton testMethods = new("Test Methods");
```
homothetyhk: Could we make it so that this button only appears outside of menu title? Maybe even in place of the test selector menu.

PR #166, opened 2026-04-10
Implement Seamstress Location

dplochcoder:
Depends on https://github.com/homothetyhk/ItemChanger.Silksong/pull/164 and https://github.com/homothetyhk/ItemChanger.Silksong/pull/165.

In addition to implementing the "Drifter's Cloak Location", this PR also adds some building blocks for future locations.
 - A basic helper for injecting previews into dialogue
 - Test helpers for setting a Bench as a starting point
 - A new QuestCompletionBool for DualLocations
   - It is feasible, though improbable, that a player marks the brolly quest completed, triggers a save, and then crashes the game before receiving the item.  This would brick a vanilla save file because the completion and item grant do not happen on the same frame, and the dialogue option to receive the reward is closed once the quest is completed.  It is probably good practice to use DualLocations in general for a variety of reasons, for instance for some future 'QuestSync' ItemSync add-on.

### DIFF
--- a/ItemChanger.Silksong/Locations/DriftersCloakLocation.cs
+++ b/ItemChanger.Silksong/Locations/DriftersCloakLocation.cs
@@ -1,0 +1,68 @@
+﻿using Benchwarp.Data;
+using HutongGames.PlayMaker;
+using HutongGames.PlayMaker.Actions;
+using ItemChanger.Locations;
+using ItemChanger.Silksong.Containers;
+using ItemChanger.Silksong.RawData;
+using Silksong.FsmUtil;
+using System.Diagnostics.CodeAnalysis;
+
+namespace ItemChanger.Silksong.Locations;
+
+public class DriftersCloakLocation : AutoLocation
+{
+    [SetsRequiredMembers]
+    public DriftersCloakLocation()
+    {
+        SceneName = SceneNames.Bone_East_Umbrella;
+        Name = LocationNames.Drifter_s_Cloak;
+    }
+
+    protected override void DoLoad()
+    {
+        this.InjectPreviewText(new("Quests", "QUEST_BROLLY_GET_DESC"), ItemChangerLanguageStrings.QUEST_BROLLY_GET_DESC_PREVIEW);
+        this.InjectPreviewText(new("Wilds", "SEAMSTRESS_BROLLY_QUEST_OFFER"), ItemChangerLanguageStrings.SEAMSTRESS_BROLLY_QUEST_OFFER_PREVIEW);
+        this.InjectPreviewText(new("Wilds", "SEAMSTRESS_BROLLY_QUEST_REOFFER"), ItemChangerLanguageStrings.SEAMSTRESS_BROLLY_QUEST_REOFFER_PREVIEW);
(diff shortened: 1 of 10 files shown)


homothetyhk:
> It is feasible, though improbable, that a player marks the brolly quest completed, triggers a save, and then crashes the game before receiving the item.

I think the "standard" way to handle this is to replace the fsm check for the conversation branch by a call to `!Placement.AllObtained` (for example, see `YarnabyLocation`). This also ensures that the location supports respawning items more or less automatically, up to ensuring that the npc is always present and able to give items.

I don't mind doing it as dual location into coordinate shiny once the quest is complete, though. I think there's even a good argument for doing that for something like Pinstress since she leaves the hut for story reasons. As a result, I'm not sure whether we should try to have a consistent implementation across NPC locations, or take them as they come and leave it to future NPC lovers/haters to normalize (e.g. to make the equivalent of HK's AvoidNPCItemDialogue).

ItemChanger.Silksong/Locations/DriftersCloakLocation.cs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    protected override void DoUnload() { }
+
+    private void ModifyFsm(PlayMakerFSM fsm)
+    {
+        fsm.MustGetState("Offer Quest").InsertMethod(0, _ => Placement?.AddVisitFlag(Enums.VisitState.Previewed));
+        fsm.MustGetState("Reoffer Quest").InsertMethod(0, _ => Placement?.AddVisitFlag(Enums.VisitState.Previewed));
+
+        // Note: It's possible for the player to brick their save if they somehow trigger a save after completing the quest but before receiving the reward.
```
homothetyhk, line 36: Is this comment still relevant? What I mean is:
- `Drifter's_Cloak` doesn't have this issue, due to the coordinate location fallback.
- `DrifterCloakLocation` alone does have this issue, but I would describe it more broadly by saying it doesn't fully implement the contract of an IC location on its own since it doesn't support respawning items.

Not an issue for the PR, but I make sure I understood the intent.

ItemChanger.Silksong/Locations/DriftersCloakLocation.cs
```diff
@@ -0,0 +1,68 @@
...
+        giveState.GetFirstActionOfType<CreateUIMsgGetItem>()!.enabled = false;
+        giveState.GetFirstActionOfType<SetFsmString>()!.enabled = false;
+        giveState.GetFirstActionOfType<SetPlayerDataVariable>()!.enabled = false;
+        giveState.AddMethod(_ =>
+        {
+            var obj = GameObject.Find("_NPCs/Seamstress");
+            Placement?.GiveAll(new()
+            {
+                FlingType = Enums.FlingType.Everywhere,
+                MessageType = Enums.MessageType.SmallPopup | Enums.MessageType.LargePopup,
```
homothetyhk, line 48: From testing, Dialog seems OK. I'll add after merging.

ItemChanger.Silksong/Locations/DriftersCloakLocation.cs
dplochcoder: The comment should probably be moved to the DualLocation definition to explain why it's a DualLocation; your analysis is accurate.

PR #168, opened 2026-04-10
Implement generic cumulative costs.

### LINKED ISSUE #87: Cumulative cost and module
Add a module for defining generic cumulative costs (using IValueProvider\<int\> and IValueProvider\<string\>). Refer to the [Hollow Knight implementation](https://github.com/homothetyhk/HollowKnight.ItemChanger/blob/master/ItemChanger/Modules/CumulativeRancidEggCost.cs) for rancid eggs.

This module is not currently planned for use in the randomizer on release, but it might be used by connections, or we might find uses that require it in the randomizer later.

dplochcoder:
Closes #87.

### DIFF
--- a/ItemChanger.Silksong/Costs/CumulativeCost.cs
+++ b/ItemChanger.Silksong/Costs/CumulativeCost.cs
@@ -1,0 +1,63 @@
+﻿using ItemChanger.Costs;
+using Newtonsoft.Json;
+
+namespace ItemChanger.Silksong.Costs;
+
+/// <summary>
+/// A cost for a resource that is tracked cumulatively, rather than being consumed at will by the player.
+/// 
+/// A shop containing two items with cumulative costs 1, 2, and 5, will show the latter two items as costing 1 and 4 after the user has purchased the first item.
+/// Likewise, after the user has purchased the second item, the final item will show as costing 3. In this way the only gating factor to purchasing any item
+/// is the total amount of resources the player has ever collected, they do not need to choose what to purchase.
+/// </summary>
+public abstract class CumulativeCost : Cost
+{
+    /// <summary>
+    /// The cumulative cost requirement for this Cost.
+    /// </summary>
+    public required int Value { get; set; }
+
+    /// <summary>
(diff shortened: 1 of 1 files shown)


ItemChanger.Silksong/Costs/CumulativeCost.cs
```diff
@@ -0,0 +1,66 @@
...
+
+/// <summary>
+/// Module for implementing cumulative costs. Typically this is implemented as an ItemChanger.Module, but this is not required.
+/// </summary>
+public interface ICumulativeCostModule
+{
+    /// <summary>
+    /// The count of the resource the player currently has available.
+    /// </summary>
+    int CurrentlyAvailable { get; set; }
```
dpinela: Should we make CurrentlyAvailable and TotalSpent writable if the only intended way to actually change them is through Spend? I'm thinking of costs based on CollectableItems in particular, where the (proper) way to add or subtract the item is with their Collect or Take methods, respectively. It's possible but awkward to bridge that with this abstraction, but what if we did the following instead?

```
int CurrentlyAvailable { get; }
int TotalSpent { get; }
void Spend(int amount);
```

That would match exactly the way the interface is used in CumulativeCost. Implementations of Spend have to do a little more work, but in return the other two properties are simpler to implement.

(This came up while implementing the Moss Druid preview tablet.)

dplochcoder: That's fine with me

ItemChanger.Silksong/Costs/CumulativeCost.cs
```diff
@@ -0,0 +1,62 @@
...
+    [JsonIgnore]
+    protected abstract ICumulativeCostModule Module { get; }
+
+    public override bool IsFree => Module.TotalSpent >= Value;
+
+    public override bool CanPay() => Module.TotalSpent + Module.CurrentlyAvailable >= Value;
+
+    /// <summary>
+    /// Returns a formatted string for the given cost amount, to be displayed in a shop context.
+    protected abstract string GetCostText(int toSpend);
```
dpinela, line 33: Missing a `</summary>` here.

PR #169, opened 2026-04-11
Implement the Pale Nails location.

### LINKED ISSUE #144: Implement Pale Nails location
This is controlled by an FSM on `Silk Needle Spell Get` in Cradle_03_Destroyed. This FSM is similar to that of crest corpses, but may be different enough that a specialized implementation is necessary.

dplochcoder:
Closes #144.

### DIFF
--- a/ItemChanger.Silksong/Locations/PaleNailsLocation.cs
+++ b/ItemChanger.Silksong/Locations/PaleNailsLocation.cs
@@ -1,0 +1,41 @@
+﻿using HutongGames.PlayMaker.Actions;
+using ItemChanger.Locations;
+using Silksong.FsmUtil;
+
+namespace ItemChanger.Silksong.Locations;
+
+public class PaleNailsLocation : AutoLocation
+{
+    protected override void DoLoad() => Using(new FsmEditGroup() { { new(SceneName!, "Silk Needle Spell Get", "Control"), ModifyFsm } });
+
+    protected override void DoUnload() { }
+
+    private void ModifyFsm(PlayMakerFSM fsm)
+    {
+        var checkState = fsm.MustGetState("Check Unlocked");
+        checkState.GetFirstActionOfType<CheckIfToolUnlocked>()?.enabled = false;
+        checkState.InsertMethod(0, _ =>
+        {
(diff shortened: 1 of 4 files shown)


ItemChanger.Silksong/Locations/PaleNailsLocation.cs
```diff
@@ -0,0 +1,41 @@
...
+        checkState.GetFirstActionOfType<CheckIfToolUnlocked>()?.enabled = false;
+        checkState.InsertMethod(0, _ =>
+        {
+            if (Placement?.AllObtained() ?? false) fsm.SendEvent("UNLOCKED");
+        });
+
+        var giveState = fsm.MustGetState("Msg");
+        giveState.GetFirstActionOfType<AutoEquipTool>()?.enabled = false;
+        giveState.GetFirstActionOfType<SpawnSkillGetMsg>()?.enabled = false;
+        giveState.AddMethod(_ =>
```
homothetyhk: This should be `.AddLambdaMethod(GiveAll)`. With AddMethod, I think the state can finish while UIDefs are waiting for input, which is not desirable. There is a virtual GetGiveInfo method which can be used to create the GiveInfo passed into GiveAll.

ItemChanger.Silksong/Locations/PaleNailsLocation.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            Placement?.GiveAll(new()
+            {
```
homothetyhk: FlingType should be the FlingType property of the location. Callback doesn't need to be set if `AddLambdaMethod` is used. MessageType.Any should be fine for this location. Container can be whatever: null, Unknown, or something descriptive like Pail Nails.

homothetyhk:
Low importance, but I'll note for future reference that:
- The respawn is set by an action on the Set Return state.
- Silk is added by the Bind Burst state.

For #153 I requested that those side effects be removed. Here, it's not an issue in the same way since this isn't a replaceable location. It might be something to revisit when we try to standardize implementations, or if we decide to move the add silk effect off of the location and onto the silk skill.

PR #171, opened 2026-04-11
Feature/issue 110 mask shards spool fragments

### LINKED ISSUE #110: Implement locations for interactable Mask Shards and Spool Fragments

kickafetus:
Closes #110 
Just learned I don't have to close and reopen PR's to track new changes. My bad


homothetyhk:
I made a few changes:
- For a few of the locations, I changed the shiny type to normal float and applied a correction to improve the shiny's visibility/positioning with respect to scenery.
- The breakable location didn't handle the slab mask shard correctly, due to that breakable not actually setting its persistent data (the breakable reforms if the room is exited without collecting the mask shard). I changed the implementation to fix the issue, and to also replace the dummy mask shard sprite.

PR #172, opened 2026-04-11
Implement Ballow location

### LINKED ISSUE #129: Implement Ballow/Diving Bell Key location
This is controlled by an FSM on `Ballow Diving Bell NPC` in Dock_12.

thomaslunn:
Adds Diving Bell Key pickup location.

Resolves #129.


ItemChanger.Silksong/Locations/BallowLocation.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            if (!Placement!.AllObtained())
+            {
+                fsm.SendEvent("FALSE");
+            }
+        });
+        
+        FsmState givenKeyState = fsm.MustGetState("Given Key?");
+        givenKeyState.RemoveActionsOfType<PlayerDataVariableTest>();
+        givenKeyState.InsertMethod(0, () =>
```
dplochcoder: nit: Could make this a ternary and save quite a few lines (fsm.SendEvent(Placement!.AllObtained() ? "TRUE" : "FALSE");)

dplochcoder (approved):

homothetyhk:
Looks great! There's one issue, which is that when the player has Silk Soar but not Everbloom, Ballow is in a state which prevents giving items (this Ballow is a BasicNPC without a dialogue FSM). I think the desired behavior overall is
- Ballow should always require `BallowMovedToDivingBell` to be in the room, regardless of Silk Soar.
- Ballow should always give items on the first interact after entering the room.
- The diving bell should always require the key to unlock, even if the player has Silk Soar and Everbloom
- The diving bell should always be usable, even if the player has Silk Soar but not Everbloom (so that it would be under repair in the base game).

The changes to the diving bell should go in a separate module, since they aren't related to just this location. It's up to you if you want to handle them in this PR or no, but I imagine removing the bell repair state is also the simplest way to fix the location issue.

thomaslunn:
Good catch on Ballow moving to repair the diving bell after obtaining Silk Soar. I've gone ahead and implemented a module for enabling the Diving Bell with any combination of abilities and without talking to Ballow and made the location dependent on it, which should handle all the mentioned behaviours.

ItemChanger.Silksong/Locations/BallowLocation.cs
thomaslunn: I've resolved this alongside other review changes for this PR. Thanks!

PR #173, opened 2026-04-12
Implement Chapel Maid location

### LINKED ISSUE #125: Implement Chapel Maid location
The Chapel Maid appears in several places, but the one who gives the Maiden's Soul is in Bonetown and controlled by an FSM on the Churchkeeper object.

thomaslunn:
Adds Chapel Maid location, with Act 3 replacement shiny.

Resolves #125.


ItemChanger.Silksong/Locations/ChapelMaidLocation.cs
```diff
@@ -0,0 +1,34 @@
...
+        FsmState snareConvoDialogState = fsm.MustGetState("Snare Soul Dlg");
+        snareConvoDialogState.RemoveActionsOfType<CollectableItemGetDataV2>();
+        snareConvoDialogState.InsertMethod(0, () =>
+        {
+            fsm.SetFsmBoolIfExists("Has Any", Placement!.AllObtained());
+        });
+
+        FsmState giveSoulState = fsm.MustGetState("Give Soul");
+        giveSoulState.RemoveActionsOfType<SavedItemGetV2>();
+        giveSoulState.InsertMethod(3, GiveAll);
```
flibber-hk:
```suggestion
        giveSoulState.InsertLambdaMethod(3, GiveAll);
```

There's a 50% chance my reading of github has led to this not being correct so please do test to make sure this doesn't break anything...

The rationale for this change is that by using the Action\<Action\> version of Insert\*Method, you don't leave the giveSoulState until after GiveAll completes. This doesn't matter now, but would matter if a big item popup were placed at this location. (You probably can't properly test the big item popups now, but when this and/or #192 are on the same branch you could try placing a few big item popups at chapel maid).

I haven't looked at the chapel maid FSM so I don't know if this is needed

flibber-hk: (I haven’t looked for this in any other prs of this type by the way, I checked this one because I was looking for an example and this seemed likely to have the issue)

thomaslunn: Thanks for the heads up, I wasn't aware of the difference between InsertMethod and InsertLambdaMethod - I think most of my more recent PRs are on InsertLambdaMethod already, but I'll give them all another look.

flibber-hk: Ah to be clear it’s not that InsertLambdaMethod is better or worse than InsertMethod but more that GiveAll(Action) is (probably) better than GiveAll() in this situation

homothetyhk:
In pre-act 3, it seems like the churchkeeper GO is destroyed when entering the room from any transition (only appears for me when loading in to respawn). I wasn't able to turn up the cause with a brief search, but I wonder if it is connected to the eventing that is supposed to move the churchkeeper to Tut_03? In any case, the location should make it so that the churchkeeper is always present in bonetown, outside of blackThreadWorld/soulSnareReady.

thomaslunn:
Thanks for the feedback - I had a quick look too and couldn't find the cause either. I'll take a deeper dive into it later, I'm somewhat busy with other things at the moment so it might take a while 🙂

dplochcoder:
> In pre-act 3, it seems like the churchkeeper GO is destroyed when entering the room from any transition (only appears for me when loading in to respawn). I wasn't able to turn up the cause with a brief search, but I wonder if it is connected to the eventing that is supposed to move the churchkeeper to Tut_03? In any case, the location should make it so that the churchkeeper is always present in bonetown, outside of blackThreadWorld/soulSnareReady.

Can you reproduce this?  She doesn't disappear for me at all.

There is an FSM on the Churchkeeper Intro Scene parent object which deactivates her when Widow is defeated, but I expect that's accounted for and you mentioned 'destroyed' instead of de-activated.

homothetyhk:
Yes, it still reproduces for me on thomaslunn:chapel-maid. The entire Churchkeeper Intro Scene object is missing/destroyed from UnityExplorer when entering the scene by means other than respawn, while running the "Chapel Maid Location" test. I remember seeing a message like \<this object has been destroyed\> in UE when reentering when I tested before, but somehow I don't see it now. Either way, it's definitely not a case of deactivating/deparenting.

dplochcoder:
> Yes, it still reproduces for me on thomaslunn:chapel-maid. The entire Churchkeeper Intro Scene object is missing/destroyed from UnityExplorer when entering the scene by means other than respawn, while running the "Chapel Maid Location" test. I remember seeing a message like <this object has been destroyed> in UE when reentering when I tested before, but somehow I don't see it now. Either way, it's definitely not a case of deactivating/deparenting.

https://github.com/homothetyhk/Silksong.Benchwarp/blob/dbf8f853a87d6dedbe9cd5d5818254c8785218f0/Benchwarp/Data/BaseGateList.cs#L799

homothetyhk:
Ah, I see, I probably should have guessed that the obstacle handler was the culprit. I'll look into fixing it--probably the change should be on the BW side to reduce the effect when using the no-save handler. Thanks for looking into this!

PR #177, opened 2026-04-13
Add Moss Druid locations

### LINKED ISSUE #124: Implement Moss Druid location(s)
This location is controlled by a FSM on `Moss Creep NPC` in Mosstown_02c.

This particular character raises some questions - should there be one location for each reward, or should they be a single multi-cost location, maybe flexible count (like Eva is in #101)? Should the rosary payouts be one of the replaceable items?

dpinela:
The Moss Druid locations come with a preview tablet, which is handled by a separate module that is added on load by any of the locations.

The mossberry costs are handled as IC costs; it is expected that either all or none of the Moss Druid locations will have items placed. This should be documented, but I'm not sure where.

This PR requires https://github.com/BadMagic100/ItemChanger.Core/issues/37 to be fixed for the preview tablet to work correctly.

Resolves #124


flibber-hk:
Is the mossberry stew location included here? (There’s a 50% chance it is and I’m blind lol :zote:)

if not, I assume it is probably in scope for the original issue

dpinela:
It isn't. I left it out because Homothety suggested making it not part of the preview tablet, and then I forgot to implement it entirely. I do think it is in scope though.

dpinela:
Updated this to reflect some of the issues that were also raised in #204.

ItemChanger.Silksong/Costs/MossberryCost.cs
```diff
@@ -0,0 +1,35 @@
+using ItemChanger.Silksong.Modules;
+using ItemChanger.Silksong.Util;
+using Newtonsoft.Json;
+
+namespace ItemChanger.Silksong.Costs;
+
+public class MossberryCost : CumulativeCost
```
flibber-hk: Flagging for whoever does the merge conflicts: can you also make this class implement IDisplayCost? 

Side note - you’ll also need to use CustomYNEnableModule.Open to open the YN box, although Homothety is probably going to move it so it might be worth waiting for that to happen before doing that change

dpinela: I did both of these, we can deal with CustomYNEnableModule.Open moving if/when Homothety does it.

homothetyhk:
I changed the preview tablet to start each placement on a new page, with the cost on its own line. It might be kind of ugly for vanilla costs, but I felt like the order that the placements are given in wasn't clear enough before. Let me know if you have any strong feelings on how it should display.

Thanks for the PR!

PR #184, opened 2026-04-14
Add PhantomLocation for Cross_Stitch (issue #138)

### LINKED ISSUE #138: Implement Phantom location
This is controlled by an FSM on Phantom in Organ_01; the vanilla Cross Stitch is given in the state `Set Data`.

kickafetus:
Hooks Phantom > Control FSM to give the randomized item on boss defeat, set defeatedPhantom, and skip the vanilla skill-get screen without fading to black or leaving the player sprite hidden.

I left in the end parry kill cutscene because it felt cool, and even in a rando I think it's still interesting?
It was actually a decent amount of work to keep that but not allow it to kick off the rest of the FSM calls for the parry skill and the screen fades and whatnot lol

Closes #138


homothetyhk:
Thanks! I love the implementation keeping the cutscene.

PR #185, opened 2026-04-14
Yn cost

### LINKED ISSUE #174: YN box
Sub-issue of #6 

It should be the case that, if a shiny is derived from a placement with a cost, a YN box should appear, and the Y option is not selectable if the cost can't be paid.

flibber-hk:
Closes #174

This PR is built on #159 because I wanted access to some of the utils added there. If I want to make changes to that PR between now and when it is merged I will probably close that PR and add the changes here (because it's more convenient for me that way)

The most important change added here that is not present there is support for costs in shiny containers. Support has also been added to tablet containers.


flibber-hk:
Draft because I have discovered regression test fails :zotesip: I'm sure I will fix them before it is my turn to be reviewed though

Issue #51, opened 2026-02-12
Bellway and Ventrica implementation details


homothetyhk:
Expected behavior:
- All bellways (resp. ventrica) can be used as sources (whether unlocked or not, whether Bell Beast is defeated or not). Travel is only allowed to unlocked bellways (resp. ventrica).
- When using a bellway/ventrica source with no unlocked targets, the menu should contain a single button which says No Available Bellway (or similar) and which closes the menu on submit.
- Bone Bottom bellway location should be collectable unconditionally. 
- Travel to Marrow bellway should place Hornet on the Bellshrine side of the room, before Bell Beast is defeated.
- Marrow bellway location should be collectable from the Bellshrine side of the room before Bell Beast is defeated.
- Bell Beast fight is not accessible from the Bellshrine side of the room. Fighting Bell Beast unlocks the ability to traverse the room and the silk heart location.
- Terminus ventrica location should be collectable from the Terminus area unconditionally.
- Bell Eater can be fought in Act 3. I think there are reasonable arguments to allow bellway use in Act 3 without fighting Bell Eater or to require fighting Bell Eater (I don't expect either choice to be difficult to support with how I expect Act 3 to be implemented); I'm open to either option.

Remarks:
-  In HK.IC, the behavior of stags was overridden entirely through StagLocation, on a per-stag basis. The IC.SS implementation can choose how much responsibility to split between locations and modules.

homothetyhk:
This is much more work than #37, so I would say it is beyond its scope.

flibber-hk:
I'm gonna work on some of the modules for this task this weekend if they haven't been done already

flibber-hk:
The totality of what I want to do here is blocked by #53 , at the moment I'm just using On.PlayMakerFSM.Start everywhere with the expectation that I'm gonna be replacing those hooks with fsm edits

PR #189, opened 2026-04-15
Implement Beastling Call location + BellEaterBypassModule

### LINKED ISSUE #145: Implement Beastling Call location
This is controlled by an FSM on `Bell Beast DefeatedCentipede NPC` (spaced like that) in Bellway_Centipede_Arena.

thomaslunn:
Beastling Call location implemented.

Bell Eater bypass module enables skipping of the Bell Eater fight in Act 3:
- Bell Beast is unaffected by the Bell Eater's presence and continues to function regardless of whether or not they have been defeated.
- All Bellway stations host entrances to Bell Eater's arena.
- The entrances stay open after Bell Eater has been defeated.

Resolves #145 and the Bell Eater portion of #51.


homothetyhk:
Looks great! I have two very low priority requests--if you want to work on them, let me know, but otherwise I can merge this and make separate issues for them.
- Bell Eater does not permanently die. Of course, there's no PD bool tracking it since the player isn't expected to return to the room, but I think it would be slightly better polish if we record its defeat to a property on the module and prevent the fight from triggering when revisiting afterwards. I'm imagining we just skip straight to Bell Beast ready to interact (as part of the module, not the location, so that it works the same whether Beastling Call is randomized or not).
- The Needolin requirement makes sense for Beastling Call, but I'm not sure it is intuitive for the boss fight. It's a very niche distinction, but maybe it could be relevant for Nuu's wish, journal entry rando, bingo, etc (stuff that cares about counting boss kills). I feel like the solution would be to have one requirement for the transition to be available, on the module (blackThreadWorld) and one requirement for the Bell Beast interact to be available, on the location (hasNeedolin). And then just put a check on the Start NPC state for Bell Beast to control whether the interact appears after the fight.

thomaslunn:
Cheers for the feedback - I will work on these tonight 🙂

thomaslunn:
This is now ready for review

PR #190, opened 2026-04-16
Implement Needle Strike location

dplochcoder:
Based on https://github.com/homothetyhk/ItemChanger.Silksong/pull/166 which introduces useful testing and preview-text utiltiies.

### DIFF
--- a/ItemChanger.Silksong/Locations/ManagedCoordinateLocation.cs
+++ b/ItemChanger.Silksong/Locations/ManagedCoordinateLocation.cs
@@ -1,0 +1,48 @@
+﻿using ItemChanger.Containers;
+using ItemChanger.Locations;
+using System.Diagnostics.CodeAnalysis;
+using UnityEngine;
+
+namespace ItemChanger.Silksong.Locations;
+
+/// <summary>
+/// Utility class for a custom location to manage alternate coordinate location spawns.
+/// </summary>
+public sealed class ManagedCoordinateLocation : CoordinateLocation, IDisposable
+{
+    private readonly Location parent;
+
+    [SetsRequiredMembers]
+    private ManagedCoordinateLocation(Location parent, Vector2 pos) : base()
+    {
+        this.parent = parent;
+
+        Name = parent.Name;
+        SceneName = parent.SceneName;
+        X = pos.x;
+        Y = pos.y;
+        FlingType = Enums.FlingType.Everywhere;
+        Managed = true;
+    }
+
+    public static ManagedCoordinateLocation Load(Location parent, Vector2 pos)
+    {
+        ManagedCoordinateLocation loc = new(parent, pos);
+        loc.LoadOnce();
+        return loc;
+    }
+
+    public void Dispose() => UnloadOnce();
+
+    public void PlaceContainer(UnityEngine.SceneManagement.Scene scene)
+    {
+        string containerType = ChooseBestContainerType();
+
+        ContainerRegistry reg = ItemChangerHost.Singleton.ContainerRegistry;
+        Container container = reg.GetContainer(containerType) ?? reg.DefaultSingleItemContainer;
+
+        PlaceContainer(
+            container,
+            ContainerInfo.FromPlacement(parent.Placement!, scene, containerType, FlingType));
+    }
+}
--- a/ItemChanger.Silksong/StartDefs/BenchwarpStartDef.cs
+++ b/ItemChanger.Silksong/StartDefs/BenchwarpStartDef.cs
@@ -1,0 +1,8 @@
+﻿using Benchwarp.Benches;
+
+namespace ItemChanger.Silksong.StartDefs;
+
+public class BenchwarpStartDef(BenchData BenchData): StartDef
+{
+    public override RespawnInfo GetRespawnInfo() => BenchData.RespawnInfo.GetRespawnInfo();
+}
--- a/ItemChangerTesting/ItemChangerTestingPlugin.cs
+++ b/ItemChangerTesting/ItemChangerTestingPlugin.cs
@@ -1,8 +1,8 @@
 ﻿using BepInEx;
 using BepInEx.Configuration;
+using ItemChanger;
 using ItemChanger.Events;
 using ItemChanger.Silksong;
-using ItemChanger.Silksong.Assets;
 using Silksong.ModMenu.Elements;
 using Silksong.ModMenu.Models;
 using Silksong.ModMenu.Plugin;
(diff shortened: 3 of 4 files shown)


ItemChanger.Silksong/Locations/NeedleStrikeLocation.cs
```diff
@@ -0,0 +1,89 @@
...
+            if (!ObtainedAny()) fsm.SendEvent("GROUND");
+        });
+
+        void PlaceRenewables(float x)
+        {
+            // If Pinstress has renewable items, place them on the ground.
+            // She might still be in the hut, but she has no 'give' dialogue to inject them into after needle strike.
+            //
+            // This could in theory be implemented as a MultiLocation, but it would be ornery to encapsulate the relevant Test in a serializable manner,
+            // since it partially depends on whether the location's items have been obtained yet or not.
```
thomaslunn: `PlacementVisitStateBool` should be exactly the test you need for the MultiLocation implementation. This is the snippet I am using for Elegy of the Deep.
```
    public static Location Elegy_of_the_Deep => new DualLocation()
    {
        SceneName = SceneNames.Tut_04,
        Name = LocationNames.Elegy_of_the_Deep,
        Test = new PlacementVisitStateBool(
            placementName: LocationNames.Elegy_of_the_Deep,
            requiredFlags: VisitState.ObtainedAnyItem,
            missingPlacementTest: null
        ),
        TrueLocation = ...
        FalseLocation = ...
    };
```
(This hasn't been merged yet - maybe there's something I'm missing!)

ItemChanger.Silksong/Extensions/ICExtensions.cs
```diff
@@ -15,6 +17,14 @@ internal static class ICExtensions
     /// Converts a struct-returning value provider to an object-returning value provider.
     /// </summary>
     public static IValueProvider<object> Embox<T>(this IValueProvider<T> t) where T : struct => new Box<T> { Source = t };
+    /// <summary>
+    /// Converts a more specific class-returning value provider to an object-returning value provider.
+    /// </summary>
+    public static IValueProvider<object> Erase<T>(this IValueProvider<T> t) where T : class => new Erased<T> { Source = t };
```
flibber-hk: Is this actually necessary? Presumably the compiler is happy to convert IValueProvider\<T\> to IValueProvider\<object\> if T is a class (so there’s no boxing involved)

dplochcoder: Apparently not.  I didn't know until now that C# differentiated between read-only and read-write covariants.

ItemChanger.Silksong/Locations/NeedleStrikeLocation.cs
dplochcoder: Unfortunately with Pinstress's placement it's more complicated than that, it would need to be multi-value enum that also takes into account Pinstress's NPC state and her Act 3 quest state.  Good to know that exists but I think the inline CoordinateLocation here is still simpler.

dplochcoder:
Somehow lost this PR for a bit; updated with the more recent utility changes.

ItemChangerTesting/LocationTests/NeedleStrikeLocationTest.cs
```diff
@@ -0,0 +1,25 @@
...
+        MenuDescription = "Tests placing items at Pinstress in the Blasted Steps.",
+        Revision = 2026041500,
+    };
+
+    public override void Setup(TestArgs args)
+    {
+        StartAt(Benchwarp.Data.BaseBenchList.Pinstress);
+        Profile.AddPlacement(Finder.GetLocation(LocationNames.Needle_Strike)!.Wrap()
+            .Add(Finder.GetItem(ItemNames.Surgeon_s_Key)!));
+        // TODO: Test persistent items when Pinstress goes away.
```
qwint: you have a test method for this :sweat_smile: what's more "TODO" in the code?

qwint:
tested:
* rosary flings, did not fling on dialogue (did on persistence) but granted rosaries without issue
* [issue]: persistence, honored persistence flag, both when set and unset, but could not respawn item *before* pinstress goes away
* tested the location while already having needle strike, worked without issue
* tested the unchecked location after forcing pinstress to leave through the quest trigger/test method, and could still successfully check the location because she did not yet leave

qwint:
just retested persistence and still had no way to get the persisted item before pinstress moves

dplochcoder:
> just retested persistence and still had no way to get the persisted item before pinstress moves

I haven't pushed a fix for that yet, this was just a rebase.  I'll reply to your original comments when fixes are uploaded, no need to retest before then.

dplochcoder:
> tested:
> 
> * multi-item placements, worked without issue
> * rosary flings, did not fling on dialogue (did on persistence) but granted rosaries without issue
> * [issue]: persistence, honored persistence flag, both when set and unset, but could not respawn item _before_ pinstress goes away

Forced the REOFFER branch so that you can always pick up persistent items again, even if Pinstress has not left yet.

> * tested the location while already having needle strike, worked without issue
> * tested the unchecked location after forcing pinstress to leave through the quest trigger/test method, and could still successfully check the location because she did not yet leave

qwint:
tested:
* rosary flings, did not fling, but no issues granting rosaries
* tested while already having needle strike, worked without issue

retested persistent items before pinstress moves, was able to grant the item again with another training bout

homothetyhk:
Regarding ad-hoc coordinate placement, I don't like the design of using unloaded locations and placements. I suspect @BadMagic100 would agree that the result of calling Get/PlaceContainer without loading is undefined. Though, maybe this is an argument to add a static method to CoordinateLocation that more clearly allows creating containers at ad hoc positions?

Alternatively, the intended way to support this is through Managed locations. In other words, 
- Store managed coordinate locations for each supported position as properties on NeedleStrikeLocation
- Load the locations when the managing location loads
- Call PlaceContainer from the fsm. The container can be selected using ChooseBestContainerType on the coordinate location, which is ok to call even if the Placement property on the location is null. Alternatively, I think it's fine for coordinate fallbacks for NPC locations to be forced shiny. Once the container is selected, the ContainerInfo is created by ContainerInfo.FromPlacement.

BadMagic100:
Yeah I would agree that most operations (other than like tagging) on an unloaded location should be considered undefined. While the managed location pattern in HK was rare I thought it was sensible and should be carried forward here if appropriate

dplochcoder:
> Regarding ad-hoc coordinate placement, I don't like the design of using unloaded locations and placements. I suspect @BadMagic100 would agree that the result of calling Get/PlaceContainer without loading is undefined. Though, maybe this is an argument to add a static method to CoordinateLocation that more clearly allows creating containers at ad hoc positions?
> 
> Alternatively, the intended way to support this is through Managed locations. In other words,
> 
> * Store managed coordinate locations for each supported position as properties on NeedleStrikeLocation
> * Load the locations when the managing location loads
> * Call PlaceContainer from the fsm. The container can be selected using ChooseBestContainerType on the coordinate location, which is ok to call even if the Placement property on the location is null. Alternatively, I think it's fine for coordinate fallbacks for NPC locations to be forced shiny. Once the container is selected, the ContainerInfo is created by ContainerInfo.FromPlacement.

Reworked the ad-hoc method into a ManagedCoordinateLocation class to make this easier (I use it 4+ times throughout these PRs).

homothetyhk:
Thanks, I like the new ManagedCoordinateLocation.

Issue #186, opened 2026-04-14
Implement fast travel locations


flibber-hk:
The fast travel objects are already being used (for the actual fast travel), so the locations should probably just be coordinate locations. The actual location is not too important, but should be near enough to the base game bellbeast/ventrica interact point that it’s obvious that the location is the fast travel location, but far enough away that the bellbeast/ventrica can be interacted with independently of the location checks. (This behaviour is the same as in HK)

The actual coordinates of the locations should be up to the person implementing this task.

The locations, with the exception of the three free fast travel locations, should be tagged with a default cost tag giving the base game rosary cost - see [the wiki](https://hollowknight.wiki/w/Fast_Travel_(Silksong)) for numbers.

This can be done independently of any open issues/prs, but obviously the costs won’t actually appear without #185 having been merged.

homothetyhk:
Fixed by #191.

PR #191, opened 2026-04-16
Implement fast travel CoordinateLocations (issue #186)

tobiaslory:
Add 19 fast travel locations (12 Bellway + 7 Ventrica) as CoordinateLocations with CostTags matching base game rosary costs. Coordinates placed near Toll Machine / tube entrance interact points in each scene. Reused coordinates from previous fast travel location test.

Update FastTravelTest to exercise all 19 locations via Finder.GetLocation(). Add FastTravelCoordinateFinder debug test for discovering object positions.

### DIFF
--- a/ItemChanger.Silksong/RawData/BaseLocationList/FastTravel.cs
+++ b/ItemChanger.Silksong/RawData/BaseLocationList/FastTravel.cs
@@ -7,6 +7,10 @@ namespace ItemChanger.Silksong.RawData;
 
 internal static partial class BaseLocationList
 {
+    // Coordinates were captured by standing Hornet at the desired check position;
+    // they reflect Hornet's transform position so containers (chests, etc.) place
+    // sensibly when this location is replaced.
+
     // === Bellway locations ===
 
     // Free — unlocked after Bell Beast
@@ -14,8 +18,8 @@ internal static partial class BaseLocationList
     {
         Name = LocationNames.Bellway__Bone_Bottom,
         SceneName = SceneNames.Bellway_01,
-        X = 52.37f,
-        Y = 21.59f,
+        X = 69.90f,
+        Y = 9.60f,
         Managed = false,
     };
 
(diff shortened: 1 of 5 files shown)


flibber-hk:
> Reused coordinates from previous fast travel location test.

fwiw these weren’t really intended to be the actual locations, if you think they’re good then that’s fine but you shouldn’t feel pressure to keep them if you think you can do better (to be honest I can’t really remember where I put them)

flibber-hk:
I went back and had a look. I think it's quite bizarre that for some of the locations the shiny is miles away from the interact point, and for other's it's right next to it. (In other words if you actually think that the previous locations - which as I said weren't should stay then you should argue why).

Also some of the new locations are probably a bit too close to the fast travel interact, for example in the image below I am standing right in front of the tube but the shiny is what wants to be interacted with. I think the shiny ought to be moved a bit further away.

image

ItemChanger.Silksong/RawData/BaseLocationList/FastTravel.cs
```diff
@@ -0,0 +1,203 @@
...
+        Managed = false,
+    };
+
+    // 40 rosaries — near Toll Machine at (83.71, 16.15)
+    public static Location Bellway__Deep_Docks => new CoordinateLocation
+    {
+        Name = LocationNames.Bellway__Deep_Docks,
+        SceneName = SceneNames.Bellway_02,
+        X = 86f,
+        Y = 16f,
```
flibber-hk: In general it should be the coordinates of Hornet that are given here (so that replacing with containers works properly in general) - i.e. the coordinates of Hornet when she's standing where the check should be (rather than the coordinates of the floor point or whatever). It might not matter for fleas in particular but it's probably good practice to get things right just in case someone comes along with another container that supports cost.

flibber-hk:
I had a look at a few of the locations, they look good to me now

PR #192, opened 2026-04-16
Accumulated features

### LINKED ISSUE #69: Implement big ui defs
The way I see it there are two general approaches that could be taken:

1) porting the big ui def from hk to Silksong, possibly with different sprites or an implementation update
2) using one/some of the prefabs in prompts_assets_all.bundle (most likely loaded via AssetHelper)

There should ideally be a way to create an arbitrary big ui def (with arbitrary text/sprites/etc); this can be done using either approach, and it should probably be down to maintainer/implementer’s choice which makes more sense. It may be the case that IC has support for an arbitrary big item popup, but also can present e.g. a custom skill or a custom crest using a modification of the relevant prompt prefab.

### LINKED ISSUE #157: Lore UI Def

### LINKED ISSUE #174: YN box

flibber-hk:
Apologies for the large PR - I didn't want to jump around between branches, and most of these features used utils from previous features hence the chaining. This PR supersedes #185, #159 and #115 (some improvements for both included here).

The most important features included:

* Big UI Defs for skills and crests (closes #69 to an extent I think is probably satisfactory, with the exception of a big UI def for focus/cast)
* Lore UI Defs and items (closes #157)
* Tablet container and locations (associated with #97 , although this is not totally complete because tablets are currently not replaceable; doing this requires adding DestroyOnContainerReplace tags and ideally measuring elevation)
* Cost support for shiny containers (closes #174)


ItemChanger.Silksong/Containers/SavedContainerItem.cs
```diff
@@ -19,13 +28,31 @@ public class SavedContainerItem : SavedItem
 
         public override void Get(bool showPopup = true)
         {
+            // If the container supports large popups, we should have the item keep control
+            // while it is giving items; large popups should not take control themselves.
+            // If not, we should not take control at all.
+            Action? callback = Callback;
+            if (RetainControl)
+            {
+                GameObject controlKeeper = new("ItemChanger Control Keeper");
```
homothetyhk, line 37: Is it worth pulling this out into something reusable (e.g. by AutoLocations)? I'm picturing a static instance component like NPCControlProxy, but for UIMsgProxy.

ItemChanger.Silksong/Containers/TabletContainer.cs
```diff
@@ -0,0 +1,118 @@
...
+    public static GameObject InstantiateWeaverTablet(Scene scene, Func<string> messageProvider)
+    {
+        GameObject tablet = GameObjectKeys.LORE_TABLET_WEAVER.InstantiateAsset(scene);
+
+        string modKey = "IC_WEAVER_TABLET";
+        LocalisedString s = new(Localization.Sheet, modKey);
+        BasicNPC npc = tablet.FindChild("Inspect Region (1)")!.GetComponent<BasicNPC>();
+        npc.StartingDialogue += () =>
+        {
+            Language._currentEntrySheets[Localization.Sheet][modKey] = messageProvider();
```
homothetyhk, line 111: I know this isn't part of the PR, but I think this is another spot where ItemChanger.Exact is useful, if I'm not mistaken.

ItemChanger.Silksong/Locations/MultiLocationEnums/FleatopiaState.cs
```diff
@@ -0,0 +1,51 @@
...
+            }
+
+            if (PlayerDataAccess.FleaGamesStarted || PlayerDataAccess.FleaGamesCanStart)
+            {
+                // Festival and PreFestival correspond to the same location, and we apply the fsm
+                // edits for both in this case
+                return FleatopiaState.Festival;
+            }
+
+            if (PlayerDataAccess.CaravanTroupeLocation == GlobalEnums.CaravanTroupeLocations.Aqueduct)
```
homothetyhk, line 43: Very minor, but this PD check is a duplicate of the first one.

ItemChanger.Silksong/RawData/LocationNames.cs
```diff
@@ -578,7 +578,6 @@ public static class LocationNames
     public const string Lore_Tablet__Cradle_Cage_3 = "Lore_Tablet-Cradle_Cage_3";
     public const string Lore_Tablet__Greymoor_Orders_Above_Home = "Lore_Tablet-Greymoor_Orders_Above_Home";
     public const string Lore_Tablet__Greymoor_Bottom = "Lore_Tablet-Greymoor_Bottom";
-    public const string Lore_Tablet__Nuu_Scroll = "Lore_Tablet-Nuu_Scroll";
```
homothetyhk, line 581: What's the rationale with the deletion here? (I'm sure that it's fine, I just want to be certain I understand it)

ItemChanger.Silksong/Locations/MultiLocationEnums/FleatopiaState.cs
flibber-hk: That's true (I did it like this because TC duplicate the PD check in the FSM) - happy to remove the second check if you prefer though

ItemChanger.Silksong/RawData/LocationNames.cs
flibber-hk: Um, you're correct to flag this because I seem to have handled this one and roach guts inconsistently. Basically my understanding is that there's no way to interact with this tablet (or roach guts) in vanilla which is why I didn't implement either of them in BaseLocationList, but I removed this one but not the roach guts from LocationNames. Not sure what the best way to proceed is

ItemChanger.Silksong/Containers/TabletContainer.cs
flibber-hk: Perhaps - in theory the code is slightly different because by using the exact sheet, you're changing the LocalisedString in the tablet rather than the return value of that localized string, which is probably slightly more hassle to do?

Another alternative would be to use a language edit rather than mutating the value in the current entry sheet but I'm not sure who'd own the language edit

homothetyhk:
I reviewed 59 out of 81 files changed (of course, leaving the hardest ones for last). I'll try to do the rest soon, though I'm traveling the next few days so it might not be right away.

ItemChanger.Silksong/Containers/TabletContainer.cs
homothetyhk: Ah, I see what you mean; exact doesn't compose well with the delegate to provide text at time of interact. I'll give it some thought.

ItemChanger.Silksong/Locations/MultiLocationEnums/FleatopiaState.cs
homothetyhk: Following the fsm is fine by me!

ItemChanger.Silksong/Containers/SavedContainerItem.cs
flibber-hk: Um - so I haven't tested with any of the AutoLocations so I don't know if it's actually necessary? In particular, it doesn't seem to be actually needed for the tablet. On the NPC AutoLocations example, I'd guess that either they just work, or they just work with slight modifications, but the UIMsgProxy.SetIsInMsg would probably neither fix the problem nor be necessary, I'd guess).

I should note that what I've named things here might not be particularly well named or placed; calling SetIsInMsg doesn't actually take control, and it doesn't actually keep control in general. What it does is set CollectableItemPickup.IsPickupPaused, and in fact according to DNSpy it's the only function which does this (hence why I used a UIMsgProxy rather than just setting it myself - it does a few other things whose importance I'm not totally sure of). 

FWIW I didn't really want to place it here, it makes more sense in the ShinyContainer, but doing that would be more work given that the moment I want to call SetIsInMsg(true) is the same as the moment the item.Get is called...

homothetyhk: I see, I didn't realize the CollectableItemPickup set was the most important part of UIMsgProxy. In that case, there probably won't be much cause to reuse.

ItemChanger.Silksong/RawData/LocationNames.cs
homothetyhk: Oh right, the fake roach guts interact. Yeah, probably both should be removed from LocationNames. I can do it when I merge.

ItemChanger.Silksong/Containers/SavedContainerItem.cs
flibber-hk: Fwiw (assuming this will be merged before the NPC PRs) we probably should remember to make sure that they either correctly support big item popups, or set their message type as SmallPopup

dpinela:
There seems to be a problem with the big UI defs - when one triggers I get an exception like this, but I don't understand how it could happen unless CopyState is doing something weird:

```
[Error  :ItemChanger.Silksong] Error on SendMessage for item Thread_Storm:
System.NullReferenceException: Object reference not set to an instance of an object
  at Silksong.FsmUtil.FsmUtil.ReplaceActionsOfType[T] (HutongGames.PlayMaker.FsmState state, System.Func`2[T,TResult] newActionGenerator) [0x00031] in /src/FsmUtil.cs:1004
  at ItemChanger.Silksong.UIDefs.BigUIDefs.MutatedDefaultBigUIDef.ApplyCustomModifications (PlayMakerFSM fsm) [0x00109] in /Users/dpinela/dev/ssmod/ItemChanger.Silksong/ItemChanger.Silksong/UIDefs/BigUIDefs/MutatedDefaultBigUIDef.cs:52
  at ItemChanger.Silksong.UIDefs.BigUIDefs.DefaultBigUIDefBase.DoSendMessage (System.Action callback) [0x00085] in /Users/dpinela/dev/ssmod/ItemChanger.Silksong/ItemChanger.Silksong/UIDefs/BigUIDefs/DefaultBigUIDefBase.cs:48
  at ItemChanger.Silksong.UIDefs.ControlRelinquishedUIDef.SendMessage (ItemChanger.Enums.MessageType type, System.Action callback) [0x0001f] in /Users/dpinela/dev/ssmod/ItemChanger.Silksong/ItemChanger.Silksong/UIDefs/ControlRelinquishedUIDef.cs:27
  at ItemChanger.Items.Item.Give (ItemChanger.Placements.Placement placement, ItemChanger.Items.GiveInfo info) [0x000d2] in /home/runner/work/ItemChanger.Core/ItemChanger.Core/ItemChanger.Core/Items/Item.cs:161
  ```

**Comment by homothetyhk** [comment]
> There seems to be a problem with the big UI defs - when one triggers I get an exception like this, but I don't understand how it could happen unless CopyState is doing something weird

I'm not able to replicate. The error line suggests a null Actions array, which would seem like an FsmUtil bug (maybe template related?).

File: ItemChanger.Silksong/Modules/YNBox/CustomYNEnableModule.cs [hunk 5]
```diff
@@ -0,0 +1,197 @@
...
+using Silksong.UnityHelper.Extensions;
+
+namespace ItemChanger.Silksong.Modules.YNBox;
+
+/// <summary>
+/// Module that allows for certain components to send more customizable YN boxes,
+/// provided that the gameObject has a <see cref="CustomYNBoxInfo"/> component.
+/// </summary>
+[SingletonModule]
+public sealed class CustomYNEnableModule : Module
```
homothetyhk, line 18: Is there a reason for `CustomYNEnableModule` to be a module over a host patch? Maybe to allow connections to globally override YN boxes/add fancy features or something like that? Of course, it's nicer organizationally to not put stuff like this in the host...

ItemChanger.Silksong/UIDefs/LoreUIDef.cs
```diff
@@ -0,0 +1,52 @@
...
+    public required LanguageString Text { get; init; }
+
+    public DialogueBox.DisplayOptions DisplayOptions { get; init; } = DialogueBox.DisplayOptions.Default;
+
+    public override void DoSendMessage(Action? callback)
+    {
+        void endConversation() => DialogueBox.EndConversation(true, callback);
+
+        DialogueBox.StartConversation(
+            Text.ToLocalisedString().ToString(false), NPCControlProxy.Instance, false, DisplayOptions, endConversation);
```
homothetyhk, line 31: Could this just be `LocalisedString.ReplaceTags(Text.Value)` with a string value provider, or do we need it for something else?

ItemChanger.Silksong/Tags/SpecialLocationTags/FleaFestivalTabletTag.cs
```diff
@@ -0,0 +1,80 @@
...
+    private IEnumerator DelaySpawnShiny(PlayMakerFSM fsm)
+    {
+        yield return new WaitForSeconds(1.5f);
+        if (fsm.ActiveStateName == "Idle Strum")
+        {
+            SpawnShiny(fsm.gameObject.transform);
+        }
+    }
+
+    // TODO - consider replacing with a call passing through container capabilities,
```
homothetyhk, line 51: I'm probably more inclined to convert to a special location and not allow replacement on this one, rather than deal with capabilities. As a future enhancement, I'd like to add a Needolin Button prompt in festival state, since it might otherwise be confusing that the tablet interact doesn't give the item.

ItemChanger.Silksong/UIDefs/LoreUIDef.cs
flibber-hk: Well the only reason we need LanguageString is to get the ReplaceTags function so yeah you're probably right

homothetyhk:
I'll go ahead and merge with this. I'm happy to make any reorganizations or other changes if the open conversations lead that way.

Thanks for all the new features!

ItemChanger.Silksong/Tags/SpecialLocationTags/FleaFestivalTabletTag.cs
flibber-hk: Sure - certainly replacement doesn't make sense because you need to play the needolin, what I meant was that if there's a container that SupportsFlingLeft that is preferred by the item(s) here, then you might want to spawn that rather than a shiny.

I *think* it still wants to be an ObjectLocation rather than an AutoLocation though, because the behaviour of the tablet during the period where all fleas have been saved but the festival is not active is essentially the same as any other (needolin) tablet. I guess having a SpecialLocation that is a subclass of ObjectLocation is still fine though? There may be another way to do it without reimplementing the tablet container logic

ItemChanger.Silksong/Modules/YNBox/CustomYNEnableModule.cs
flibber-hk: > it's nicer organizationally to not put stuff like this in the host

This is the main reason I did it like this; I think if connections wanted to override YN boxes it'd probably make more sense to hook various overloads of DialogueYNBox.Open. (And also currently the host is only Harmony patches, but the `CollectableItemPickup.Pickup.ControlFlowPrefixMoveNext` hook really wants to be MonoDetour so I didn't want to think about how to organize that :zote:)

One reason to move this to the host is that in some sense this module is "more default" than other modules - for example the baldur shell reduced HP module from HK.IC is default because the randomizer needs it, but it's plausible that a light IC client might want it to be absent. Whereas there's no compelling reason to explicitly want CustomYNEnable module to be absent. (This is probably the case for most/all default modules in SilksongHost.)

I'd say this is a decision that should probably be made, either:
- Modules are only allowed to be default modules if they are maximally non-objectionable, and any modules like the equivalent of the baldur shell HP module have to be explicitly added by rando
- Modules (like the baldur shell HP module) are allowed to be default modules if rando always wants them and there's a compelling case that they should be used most of the time, but anything like the CustomYNEnable module is moved to the host
- IC.Core is changed so that there are multiple tiers of default module, and the person creating the profile gets to choose whether the "must have" tier is created or whether the "nice to have" tier is created.

(I think I favour option 1, although that's not exclusive with this being moved to the host)

flibber-hk:
> There seems to be a problem with the big UI defs - when one triggers I get an exception like this, but I don't understand how it could happen unless CopyState is doing something weird

I too haven't ever seen this. Do you see the big item popup? Are your FSMUtil and AssetHelper up to date?

(I assume you tested with the crest big ui def in the greyroot branch? That uses the component rather than the fsm to control the popup so shouldn't be affected)

flibber-hk:
> I'm happy to make any reorganizations or other changes if the open conversations lead that way.

I didn't want to bump IC.Core because of potential conflicts with other PRs, but the latest version of IC.Core seems to have resolved issues with the container info component so you should remember to remove the one I put in the tablet container when you get round to doing that

ItemChanger.Silksong/Modules/YNBox/CustomYNEnableModule.cs
flibber-hk: On second thought another way to organize it would be to change the module to be a YNUtil class or whatever, such that in a static Init function it subscribes to OnEnterGame and OnLeaveGame (or maybe OnItemchangerHook/Unhook)

homothetyhk: Yeah, I think the *Util pattern makes sense for this sort of thing. To me, the test for whether something should be a module or not comes down to configurability. For example, `ConsistentRandomnessModule` needs to be a module to store its seed; `ObstacleHandlerModule` needs to be a module because obstacle handling is highly opinionated and needs lots of layers of configurability. For modules which just carry a small hook to support a custom component and things of that nature, I think it makes more sense to let the host manage the hook.

We could go with Option 3 by storing the list of which *Util classes to load on a default module. It might be a little excessive in terms of layers of indirection, but it would act to bundle essential features together to simplify building up a module list from scratch. But maybe that isn't necessary until someone asks to be able to disable them.

PR #195, opened 2026-04-18
Implement the Everbloom location.

dplochcoder:
Plenty of opportunity to add preview text here, but like the Black Egg Temple I don't think this location needs it.  Implemented as a DualLocation so the player need only traverse the red memory once.

### DIFF
--- a/ItemChanger.Silksong/Locations/EverbloomLocation.cs
+++ b/ItemChanger.Silksong/Locations/EverbloomLocation.cs
@@ -1,0 +1,24 @@
+﻿using Benchwarp.Data;
+using HutongGames.PlayMaker.Actions;
+using ItemChanger.Locations;
+using Silksong.FsmUtil;
+
+namespace ItemChanger.Silksong.Locations;
+
+public class EverbloomLocation : AutoLocation
+{
+    protected override void DoLoad() => Using(new FsmEditGroup() { { new(SceneName!, PrimitiveGateNames.door_wakeInRedMemory_Root, "Wake Up"), ModifyFsm } });
+
+    protected override void DoUnload() { }
+
+    private void ModifyFsm(PlayMakerFSM fsm)
+    {
+        // Don't give everbloom.
+        fsm.MustGetState("Set State").GetFirstActionOfType<SavedItemGetV2>()?.enabled = false;
+
+        var giveState = fsm.MustGetState("Get Item Msg");
+        giveState.GetFirstActionOfType<CreateUIMsgGetItem>()?.enabled = false;
+        giveState.GetFirstActionOfType<SetFsmString>()?.enabled = false;
+        giveState.AddMethod(_ => GiveAll(() => fsm.SendEvent("GET ITEM MSG END")));
+    }
+}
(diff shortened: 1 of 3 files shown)


ItemChanger.Silksong/Locations/EverbloomLocation.cs
```diff
@@ -0,0 +1,20 @@
...
+
+public class EverbloomLocation : AutoLocation
+{
+    protected override void DoLoad() => Using(new FsmEditGroup() { { new(SceneName!, "door_wakeInRedMemory Root", "Wake Up"), ModifyFsm } });
+
+    protected override void DoUnload() { }
+
+    private void ModifyFsm(PlayMakerFSM fsm)
+    {
+        var giveState = fsm.MustGetState("Get Item Msg");
```
flibber-hk: How does the base game FSM give the everbloom item? It seems weird that there's nothing here to remove it

(EDIT: I believe the answer is on the previous state)

dplochcoder: I missed it, thanks.  I don't know why I thought the CreateUIMsgGetItem did it when it never does elsewhere.

ItemChanger.Silksong/Locations/EverbloomLocation.cs
```diff
@@ -0,0 +1,23 @@
+﻿using HutongGames.PlayMaker.Actions;
+using ItemChanger.Locations;
+using Silksong.FsmUtil;
+
+namespace ItemChanger.Silksong.Locations;
+
+public class EverbloomLocation : AutoLocation
+{
+    protected override void DoLoad() => Using(new FsmEditGroup() { { new(SceneName!, "door_wakeInRedMemory Root", "Wake Up"), ModifyFsm } });
```
flibber-hk: Do you think this would be better as `PrimitiveGateNames.door_wakeInRedMemory_Root` ?

dplochcoder: Sure, didn't know those constants existed.

homothetyhk:
Looks great!

One issue: the flingType on the EverbloomLocation instance should be DirectDeposit. However, even with DirectDeposit, a RosariesItem doesn't seem to give any rosaries from the location. I confirmed that HeroController.AddCurrency was called, so it might be something specific to the memory scene or having HUD set to OUT.

dplochcoder:
> Looks great!
> 
> One issue: the flingType on the EverbloomLocation instance should be DirectDeposit. However, even with DirectDeposit, a RosariesItem doesn't seem to give any rosaries from the location. I confirmed that HeroController.AddCurrency was called, so it might be something specific to the memory scene or having HUD set to OUT.

I'll address this in a separate PR, should be a straightforward fix.  Fixed the fling issue and added a test setup.

dplochcoder:
> Looks great!
> 
> One issue: the flingType on the EverbloomLocation instance should be DirectDeposit. However, even with DirectDeposit, a RosariesItem doesn't seem to give any rosaries from the location. I confirmed that HeroController.AddCurrency was called, so it might be something specific to the memory scene or having HUD set to OUT.

This turned out to not be as straightforward as I hoped.  I've documented my findings in https://github.com/homothetyhk/ItemChanger.Silksong/issues/247.

PR #197, opened 2026-04-19
Add Sprintmaster Swift race locations and skip module (#131)

### LINKED ISSUE #131: Implement Sprintmaster Swift location(s)
Swift's rewards are controlled indirectly through the Behaviour FSM on `Sprintmaster Runner` in Sprintmaster_Cave.

The game uses the SprintRaceController attached to each race track object to store the reward for each race, but for IC purposes we probably want to bypass that anyway.

kickafetus:
Closes #131


ItemChanger.Silksong/RawData/LocationNames.cs
```diff
@@ -450,6 +450,7 @@ public static class LocationNames
     public const string Rosary_String__Far_Fields_Seamstress = "Rosary_String-Far_Fields_Seamstress";
     public const string Rosary_String__Shellwood_Bench = "Rosary_String-Shellwood_Bench";
     public const string Rosary_String__Shellwood_Chapel = "Rosary_String-Shellwood_Chapel";
+    public const string Rosary_String__Sprintmaster_Race_1 = "Rosary_String-Sprintmaster_Race_1";  // Race 1 reward (skip path only)
```
dplochcoder: Also add the relevant BaseLocationList properties?

ItemChangerTesting/LocationTests/SprintmasterLocationTest.cs
```diff
@@ -0,0 +1,142 @@
...
+        MenuName = "Sprintmaster - Quest Completion (Mask Shard)",
+        MenuDescription = "Hooks final quest Mask Shard reward. Complete all races to receive IC item instead.",
+        Revision = 2026041600,
+    };
+
+    public override void Setup(TestArgs args)
+    {
+        StartAt(new CoordinateStartDef() { SceneName = "Sprintmaster_Cave", X = 60.82f, Y = 8.57f, MapZone = GlobalEnums.MapZone.NONE });
+
+        Profile.AddPlacement(new SprintmasterLocation
```
dplochcoder: This should be obtained through Finder.

ItemChanger.Silksong/Modules/SprintmasterSkipModule.cs
```diff
@@ -0,0 +1,174 @@
...
+        FsmState endDialogue3 = fsm.MustGetState("End Dialogue 3");
+        endDialogue3.InsertMethod(0, () =>
+        {
+            if (!skipApplied) return;
+            skipApplied = false;
+
+            // Race 1: IC placement if registered, vanilla otherwise.
+            if (ActiveProfile != null
+                && ActiveProfile.TryGetPlacement(LocationNames.Rosary_String__Sprintmaster_Race_1, out Placement? race1Placement)
+                && !race1Placement!.AllObtained())
```
dplochcoder: This check should be removed, here and for race 2.  We should not grant the vanilla reward if a placement exists at all (it may already be obtained through e.g. ItemSync)

ItemChanger.Silksong/Modules/SprintmasterSkipModule.cs
```diff
@@ -0,0 +1,174 @@
...
+            int lastRaceIdx = arrayGetAction.array.Length - 1;
+
+            race1Reward ??= (arrayGetAction.array.Get(0) as GameObject)
+                ?.GetComponent<SprintRaceController>()?.Reward;
+
+            // Pre-increment the quest counter for each skipped race so Was Extra? routes to Quest End.
+            FullQuestBase? quest = questCheckAction?.Quest.Value as FullQuestBase;
+            if (quest != null)
+            {
+                for (int i = 0; i < lastRaceIdx; i++)
```
dplochcoder: Is 0 the correct starting index here?  What happens if the user does 1 race, then decides on the second prompt to skip to the 3rd?

ItemChanger.Silksong/Locations/SprintmasterLocation.cs
```diff
@@ -0,0 +1,91 @@
...
+                if (rewardItem != null && !Placement!.AllObtained())
+                    GiveAll();
+            }
+        });
+    }
+
+    private void HookRace2BeastShard(PlayMakerFSM fsm)
+    {
+        // Give Reward → SavedItemGet gives the current race's reward.
+        // Race 2 is identified by SavedItem name "Great Shard"; Race 1 (Rosary String) passes through.
```
dplochcoder: Are we not also defining a placement for the Race 1 reward?  Why not?
It seems it would be easier here to specify the race index this location represents, than to compare to vanilla reward strings.  This would presumably also be a more extensible pattern for if we ever randomize the memento.

ItemChanger.Silksong/Modules/SprintmasterSkipModule.cs
```diff
@@ -0,0 +1,174 @@
...
+/// Race 1 Rosary String (IC placement if registered, vanilla otherwise),
+/// Race 2 Beast Shard (IC placement if registered), and Race 3 Mask Shard (via SprintmasterLocation).
+/// </summary>
+public class SprintmasterSkipModule : Module
+{
+    // skipActive: set on yes, cleared when Array Track? applies the jump.
+    private bool skipActive;
+    // skipApplied: set after the jump, cleared when End Dialogue 3 gives bypassed rewards.
+    private bool skipApplied;
+    // Race 1 vanilla reward (Rosary String), captured when the skip is applied.
```
dplochcoder: Should we also capture the race 2 reward?

dplochcoder (changes requested):

kickafetus:
A lot of this is my bad with forgetting about the main concerns around getting the location setup.
I was too focused on the skip module working and giving all three with a dialogue pop up.

I will be resolving what you have brought up, here very shortly. Thank you very much for calling this all out.
Also just forgot about the memento as well, and will be fully baking that in too.

I will address each point you bring up above, but wanted to give this general response here first... my b

ItemChangerTesting/LocationTests/SprintmasterLocationTest.cs
kickafetus: setup through finder

ItemChanger.Silksong/RawData/LocationNames.cs
kickafetus: So I did setup a Sprintmaster baselocation list, but should I actually setup and use a Quest baselocation list?

ItemChanger.Silksong/Modules/SprintmasterSkipModule.cs
kickafetus: fixed

ItemChanger.Silksong/Modules/SprintmasterSkipModule.cs
kickafetus: Stopped using the direct index, 0, for starting and pulled currentRaceIdx
Tested with saying no to the dialogue popup, doing race 1, and then skipping to the end.
Didn't get a duplicate and it incremented correctly

ItemChanger.Silksong/Locations/SprintmasterLocation.cs
kickafetus: Specified the placements for race 1, 2, and the memento for the extra race.
+Memento added with check for the white flower

ItemChanger.Silksong/Modules/SprintmasterSkipModule.cs
kickafetus: also setup the capture of the race 2 reward

ItemChanger.Silksong/Modules/SprintmasterSkipModule.cs
```diff
@@ -0,0 +1,168 @@
...
+using Silksong.FsmUtil;
+
+namespace ItemChanger.Silksong.Modules;
+
+/// <summary>
+/// Inserts a yes/no skip prompt at the start of the Sprintmaster conversation.
+/// Choosing yes jumps to the final race and awards all bypassed rewards on quest completion:
+/// Race 1 Rosary String (IC placement if registered, vanilla otherwise),
+/// Race 2 Beast Shard (IC placement if registered, vanilla otherwise),
+/// and Race 3 Mask Shard (via SprintmasterLocation).
```
dplochcoder, line 19: With the memento possibly randomized as well, I think this gets a little messier.  The player will likely only want to skip to the memento race if it is randomized, and this may depend on their gear, etc.

I think a clean solution would be to change the singular y/n prompt into a sequential set of prompts, i.e. "Skip Race 1? y/n" into "Skip Race 2? y/n" into "Skip Race 3? y/n" iff the player has Everbloom and is allowed to run the memento race.  WDYT?

ItemChanger.Silksong/RawData/BaseLocationList/Sprintmaster.cs
dplochcoder, line 1: Generally locations should be organized by item pool, not location name; this will be crucial later when implementing the randomizer mod.

This will require merging later because there are a number of pending PRs adding new locations, but I think these 4 locations would in order go to files named:
 - Rosaries.cs
 - ShellShards.cs
 - MaskShards.cs
 - Mementos.cs

ItemChanger.Silksong/RawData/LocationNames.cs
dplochcoder: See my comment on Sprintmaster.cs; locations should generally be organized by item pool.

ItemChangerTesting/LocationTests/SprintmasterLocationTest.cs
dplochcoder, line 1: This is a lot of tests for one 'location' with a lot of duplicated setup.  Can we combine these into a single test?  The player can check each race / the skip module by simply redoing the test a different way.

ItemChangerTesting/ModuleTests/SprintmasterMementoFsmLoggerModule.cs
dplochcoder, line 1: This isn't a module test; this should probably be a file-visible class in SprintmasterLocationTest.cs

dplochcoder (changes requested):

ItemChanger.Silksong/RawData/BaseLocationList/Sprintmaster.cs
kickafetus: Makes sense.
I will go ahead and split these out

ItemChanger.Silksong/Modules/SprintmasterSkipModule.cs
kickafetus: So at the moment, I actually only have it jumping to Race 3, not to the extra 4th race. Apologies for the outdated comment., I will resolve that as well when I tackle these few extra things below.
It is still the vanilla behaviour of requiring you to have done the first three races before attempting the extra race.

In the moment this still seemed like the easiest solution, as it was still baked into the normal game checks.
I personally feel like that would be too many dialogue popups for something like this, but I still like the idea of running a check to see if the player has the white flower when you approach.
Skip to race 3? y/n
-check to see white flower-
Skip further to the Memento Race? y/n

Thoughts on that part?
I'm happy working it out either way, though.

ItemChangerTesting/LocationTests/SprintmasterLocationTest.cs
kickafetus: I do need to limit how many tests I throw into my PRs, thanks for the call out.
I'll just leave the last one with the white flower present, and like you said, just let testers run through the races as they see fit.

ItemChanger.Silksong/Modules/SprintmasterSkipModule.cs
kickafetus: So since I have also setup a preview for items for the wishwalls, what are your thoughts on us showing what the race rewards are?
I could see it as an extension of the QoL for the option of skipping through it.

Totally would be extra, but curious of your opinion as well, as the thought popped into my head

ItemChangerTesting/LocationTests/SprintmasterLocationTest.cs
dplochcoder: It's mostly a UI issue, since the tests have to navigated via scroll in the mod menu.  It would be less of an issue if the menu had a text filter or a wider interface.

ItemChangerTesting/ModuleTests/SprintmasterMementoFsmLoggerModule.cs
kickafetus: Just removed it since we shouldn't need that logger anyway once this is live

ItemChanger.Silksong/Modules/SprintmasterSkipModule.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                Finish();
+                return;
+            }
+
+            DialogueYesNoBox.Open(
+                yes: () => { onYes(); Finish(); },
+                no: () => Finish(),
+                returnHud: true,
+                text: "Skip to the final race and collect all Sprintmaster rewards?"
```
flibber-hk: FWIW text displayed to the user (under normal circumstances) shouldn't be written in code - this message should appear in default.json and probably be retrieved using the same mechanism as with the other language strings in `RawData.ItemChangerLanguageStrings.cs`

ItemChanger.Silksong/Modules/SprintmasterSkipModule.cs
dplochcoder: Re https://github.com/homothetyhk/ItemChanger.Silksong/pull/197#discussion_r3133666028 - Yea that seems like a nice middle ground to max out at 2 y/ns instead of possibly 3.  I think the 'skip to memento race' should only be offered if the memento placement is present also but that's minor.

I feel like with the skip module (and even without it) a preview for Sprintmaster Swift rewards is kind of overkill.  Generally previews exist to save the player spending time going after long quests with rewards they don't want, but in this case (a) they must've already made it to the cave which is a huge trek and a form of late game on its own, and (b) doing the race takes about a minute or two, there's no farming or questing of any kind.

ItemChanger.Silksong/Locations/SprintmasterLocation.cs
```diff
@@ -0,0 +1,126 @@
...
+
+    private void HookSprintmaster(PlayMakerFSM fsm)
+    {
+        if (IsQuestCompletion)
+            HookQuestCompletion(fsm);
+        else
+            HookRaceReward(fsm);
+    }
+
+    private void HookQuestCompletion(PlayMakerFSM fsm)
```
dplochcoder, line 49: TODO: We should probably define a generic QuestRewardLocation to use here instead.  There are other places, like Throwing_Rings, that give the quest reward directly through a dialogue FSM, that could be implemented by such a location.  (Doesn't need to be done here, PR is fine without it I think.)

ItemChangerTesting/LocationTests/SprintmasterLocationTest.cs
```diff
@@ -0,0 +1,47 @@
+using ItemChanger.Silksong.Modules;
+using ItemChanger.Silksong.RawData;
+using ItemChanger.Silksong.StartDefs;
+
+namespace ItemChangerTesting.LocationTests;
+
+internal class SprintmasterMementoLocationTest : Test
```
dplochcoder: nit: Rename to SprintmasterLocationTest to match file?  Also update metadata

dplochcoder (approved):

ItemChanger.Silksong/Locations/SprintmasterLocation.cs
flibber-hk: That's presumably some minor variation (if any) of the Wishwall location implied by #162 right?

dplochcoder: Yea, I think ideally WishwallLocation would extend QuestRewardLocation, you could extract the quest reward bits from the quest board parts.

kickafetus: The wishwall is a replacement of the normal item and just let the game give me that item, where as in here, it replaces the FSM call.

Tbh I haven't looked at the wishwall implementation since the last interaction on it, but I am curious if I should go back and replace it with some FSM replace since I am more experienced with logging and pulling info on FSM's or if we should always do the proxy item.

Thoughts?

kickafetus: If we want to go the route of it always doing the proxy item, I can setup that module.
Is this how we would probably want to advise people to go about quest completion rewards then, when it comes to NPC's?

ItemChangerTesting/LocationTests/SprintmasterLocationTest.cs
kickafetus: Haven't done this yet, but I will make sure to get back to it

ItemChanger.Silksong/Modules/SprintmasterSkipModule.cs
kickafetus: Thanks for the heads up and direction here!
I have put the new text into default.json, but still need to figure out exactly what I want them to say.

I know it can be simple, but I want to put a little thought into it since it will be end-user facing, and I want to match what the game usually gives.

ItemChanger.Silksong/Locations/SprintmasterLocation.cs
flibber-hk: I don’t think there’s any need to make any changes along these lines right now (but I haven’t properly thought about the wishwall location pr so make of that what you will)

qwint:
did not fully test, but found some issues 
* when granted alone the crest uidef from beast crest overlapped with dialogue,
* after losing a race you can be prompted for a rematch but specifically for the skip to race 4 it seems to drop back to race 3 instead (seems to honor race 4 if you actually complete race 3 properly). Also while completing this faux race 3 does grant the 1-3 rewards, it seems to put you in some state where you can no longer trigger the lap gate in race 4 making that location inaccessible (if not completing the faux race 3, then failing it)
* I also was not able to easily reproduce, but some mix of refusing and taking the skips/rematches put me in a state where the race 1-2 rewards were not granted but I could only re-do race 4 making 1 and 2 inaccessible

Also re the skip design, is it intentional that you have a skip to 3 button, then an additional skip to 4 button, but no way to skip to 2?
I would assume either you can skip to any you want or you could skip directly to the end
additionally is it intentional that once you reach race 4 you are still prompted asking if you want to skip to race 4?

PR #198, opened 2026-04-19
Implement slab kidnapping module

### LINKED ISSUE #88: Module to manage the Slab kidnapping sequence
Refer to the discussion in the [planning issue](https://github.com/homothetyhk/Silksong-Randomizer-Organization-Stub/issues/9). The module should change the sequence to be available according to an `IValueProvider<bool>` property of the module, which by default should be constant true. The module should also prevent the game from removing Hornet's items, and should apply any fixes (e.g. to the Slab arena) needed to account for this.

Optionally, there could be a property on the module configuring whether Hornet's items are removed (defaulting to no). I don't know whether the method of removing items would need to change to cooperate with other IC features like cursed/cloakless crest.

thomaslunn:
Adds a module for controlling the Slab kidnapping sequence.

While the module is loaded, the player's items are not taken when being kidnapped by a wardenfly. The module contains an IValueProvider controlling whether jailers are available in each of the 3 kidnapping spots, defaulting to always true. There's an exception for the Greymoor jailer, which is not present when the Moorwing boss fight is happening in the same room - I could see us going either way on this one, see the [planning issue.](https://github.com/homothetyhk/Silksong-Randomizer-Organization-Stub/issues/9)

_Update: Being captured while cursed can now optionally transport you to The Slab._

Resolves #88.


ItemChanger.Silksong/Modules/SlabKidnappingModule.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                new(SceneNames.Greymoor_05, "Slab Fly Large Cage", "Control"),
+                fsm => HookWardenfly(fsm, () => PlayerDataAccess.CurseKilledFlyGreymoor)
+            },
+        });
+    }
+
+    protected override void DoUnload()
+    {
+        ItemChangerHost.Singleton.GameEvents.AddSceneEdit(SceneNames.Bone_East_04c, ForceJailerDocks);
```
dplochcoder: This and below should be `RemoveSceneEdit`.  Alternatively, consider using [SceneEditGroup](https://github.com/homothetyhk/ItemChanger.Silksong/pull/178/changes/36ec1eef0588582f9b33d6f747dd65b9e1353bcd) when it lands.

ItemChanger.Silksong/Modules/SlabKidnappingModule.cs
```diff
@@ -0,0 +1,168 @@
...
+
+    protected override void DoLoad()
+    {
+        ItemChangerHost.Singleton.GameEvents.AddSceneEdit(SceneNames.Bone_East_04c, ForceJailerDocks);
+        ItemChangerHost.Singleton.GameEvents.AddSceneEdit(SceneNames.Shadow_21, ForceJailerBilewater);
+        ItemChangerHost.Singleton.GameEvents.AddSceneEdit(SceneNames.Greymoor_05, ForceJailerGreymoor);
+
+        Using(new FsmEditGroup()
+        {
+            {
```
dplochcoder: Might as well use a wildcard for this, unless there are "Slab Fly Large Cage"s you explicitly don't want to target?

dplochcoder (changes requested):

ItemChanger.Silksong/Modules/SlabKidnappingModule.cs
thomaslunn: Good catch, thanks. I am eagerly awaiting the merge for SceneEditGroup!

ItemChanger.Silksong/Modules/SlabKidnappingModule.cs
thomaslunn: I don't believe I can without losing functionality - the state of whether the jailer has been killed by curse is tracked by a different PlayerData bool in each location, so the `isWardenflyDefeatedByCurse` argument needs to be different for each. This is only used to spawn the appropriate background objects when the jailer is not present, so not a particularly essential feature.

thomaslunn: In hindsight, this would be clearer if the argument type was `IValueProvider<bool>` rather than `Func<bool>` - then I could instantiate a `PDBool` for each one.

dplochcoder (approved):

ItemChanger.Silksong/Modules/SlabKidnappingModule.cs
dplochcoder: Could you also not grab the pdbool name from the fsm, instead of passing it in?  Assuming it's in the same place in all of them

dplochcoder: Actually, re-evaluating here - the point of this module is to ensure the slab kidnapping is always available if the provider says so, so should it even respect the killedByCurse bools at all?  I think the answer is 'no', the fsm should simply ignore the curse and skip the curse-kill part instead.

thomaslunn: Ok, for context, the only scenario where those bools are used at the moment is in the following situation:

1. Wardenflies are available (Provider is true)
2. A wardenfly is killed by the curse - this doesn't permanently remove the wardenfly from the scene
3. Later, wardenflies are no longer available (Provider is false)

In specifically this situation, the bools are used to place the curse background objects in the scene. In our current expected rando implementation this should never come up since we intend to have the wardenflies always present, but in a world with different provider settings it could realistically happen.

To clarify, are you suggesting that the wardenflies being killed by the curse should be removed entirely? IE the exception is removed and the player is transported to the slab when captured while cursed. I'm not particularly attached to the kill sequence so I could see that being an acceptable QOL which would also simplify the code somewhat.

dplochcoder: > the exception is removed and the player is transported to the slab when captured while cursed.

Yes, otherwise it may be possible for a rando save to hard lock if Slab Capture is intended progression but the player winds up cursed first (and is unable to break the curse).

thomaslunn: Ok, I have made the change, and removed the player bool checks :slightly_smiling_face:

flibber-hk: > (and is unable to break the curse)

I don't think this is actually a thing - being unable to break the curse would lead to more problems than just not being able to follow the slab kidnap... I believe the behaviour Homothety has described is that the cursed crest can be unequipped, but you have to carry the curse from a bench to Yarnaby's office if you want to get the witch's crest check.

I think my expectation would be that if you're cursed then it does kill the wardenfly (as in the base game), but it doesn't save the fact that the wardenfly has been killed (so you can't actually lock yourself out of the slab kidnap).

Probably worth getting a third or fourth opinion before making further changes though...

thomaslunn: After discussion in Discord, I've gone with cursed crest preventing slab capture by default, controllable via a ValueProvider on the module. Thanks both of you for the input!

PR #202, opened 2026-04-20
Implement Shakra Locations

dplochcoder:
This PR does a lot of things to make Shakra work safely and interestingly for rando, in addition to defining all Shakra locations.

-   ShopPlacements are refactored into ShopMultiPlacement (the default) and ShopFlexiblePlacement (currently only used by Shakra).  The latter provides a way to put multiple Items into a single shop purchase, which allows full-flexible to work with the new ShakraMapLocations without multiplying the default location cost.  Because ShopFlexiblePlacement is an ISingleCostPlacement, rando does not need to specify the rosary cost either, it can be automatically applied by IC.
-   ShakraModule implements all scene and FSM modifications, ensuring that Shakra can be permanently found at all map discovery locations.  ShakraModule gets installed if any ShakraShopLocation is included.
    -   Shakra will only appear in Bonebottom after Bell Beast is defeated, but the Mosslands map is always available regardless.
    -   Shakra will only appear in Bellhart after Widow is defeated, but the Bellhart map is available as soon as the player has visited Bellhart.
    -   Shakra will disappear from both of these locations in Act 3 to go defend Bellhart in Greymoor, but will remain at all the non-town locations.
    -   Shakra otherwise doesn't disappear when the Trail's End quest becomes available.
-   Encountering Shakra in the Hunter's March arena is optional, and will grant access to the Hunter's March Map but this can never be required because the player might visit Greymoor, Bellhart, or Shellwood first and miss the encounter.  Thus, Shakra will always be present in Ant_20 regardless of her presence in the arena.
-   QoL: Shakra's inventory is sorted to always show the item for the on-location Map placement first, then the rest of the discovered map placements, then Shakra-Global placements, then any vanilla inventory.  Thus whenever talking to Shakra somewhere new, the first item is the new item.  Players can buy passed-on items anywhere at any later point allowing for interesting spending decisions.


ItemChanger.Silksong/Extensions/PlayerDataTestExtensions.cs
```diff
@@ -26,4 +28,20 @@ public static void Modify(this PlayerDataTest pdt, Func<PlayerDataTest.Test, Pla
...
+                Type = TestType.Bool,
+                FieldName = ALWAYS_TRUE,
+                BoolValue = result
+            }]
+        }];
+    }
+
+    private const string ALWAYS_TRUE = "ITEMCHANGER_ALWAYS_TRUE";
+
+    static PlayerDataTestExtensions() => PrepatcherPlugin.PlayerDataVariableEvents<bool>.OnGetVariable += (_, name, orig) => name == ALWAYS_TRUE || orig;
```
flibber-hk: I believe it would be more idiomatic for this hook to be profile scoped (rather than persistent)? Most likely this would be done by the host, in [similar-ish cases](https://github.com/homothetyhk/ItemChanger.Silksong/blob/81062f609a5f65ebae730a603d21a2ddfd760cbf/ItemChanger.Silksong/Serialization/AtlasSprite.cs#L78) I've had persistent subscriptions to the host's lifecycle events

dplochcoder: Seems messy to do that for a niche utility that may not ever be invoked; if I switch this to ALWAYS_FALSE I can just avoid the hook instead.

ItemChanger.Silksong/Extensions/ICExtensions.cs
```diff
@@ -83,6 +129,26 @@ public static bool TryToPay(this Cost c)
...
+    /// <summary>
+    /// Returns all sub-costs of this possible Multicost.
+    /// </summary>
+    public static IEnumerable<Cost> Flatten(this Cost cost)
+    {
+        if (cost is MultiCost multi)
+        {
+            foreach (var c1 in multi)
+            {
+                foreach (var c2 in Flatten(c1)) yield return c2;
```
flibber-hk: Not sure if this actually needs changing but MultiCosts get flattened in the [constructor](https://github.com/BadMagic100/ItemChanger.Core/blob/a15387ccd95a6c9cf06624fb0d9e80e49830f4e8/ItemChanger.Core/Costs/MultiCost.cs#L52), so you could simply yield return c1

ItemChanger.Silksong/Costs/CollectableItemCost.cs
```diff
@@ -0,0 +1,51 @@
+﻿using ItemChanger.Costs;
+using ItemChanger.Silksong.Extensions;
+using Newtonsoft.Json;
+
+namespace ItemChanger.Silksong.Costs;
+
+public class CollectableItemCost : Cost
```
flibber-hk: Can you make this implement IDisplayCost? And ShellShardCost implement IDisplayCost and ICurrencyCost? (That way they render properly in YN boxes, assuming they get opened with the mechanism I added in my big PR)

dplochcoder:
> * As I reported about Forge Daughter and 12th Architect in [Implement Shops #178](https://github.com/homothetyhk/ItemChanger.Silksong/pull/178), Shakra also displays her unmodded inventory when benchwarping directly to a room she's in. However, this only seems to happen with standing instances of Shakra, and only when she speaks to Hornet automatically upon getting near her for the first time. In practice, there aren't any standing Shakras in rooms with benches except for Mt. Fay, and there the bench is inside her trigger* so it would normally be impossible to have unlocked that bench without having triggered that first meeting event.

I believe this is the same root cause, I can't reproduce this anymore.

> * Even without benchwarping directly to the room, standing Shakra won't set the flag for having spoken to her at a given location from the first meeting event, so when she opens her inventory the item that's supposed to be at that location won't be there until after speaking to her in that location manually (tested with Deep Docks and Mt. Fay).

Fixed the FSM edit to just check the shop opening state, which AFAICT is always eventually reached from any conversation entry point.

> * At Shakra's Mt. Fay location, having Faydown Cloak causes her lantern to be visually missing. The warmth zone is still present, but invisible.

Added a tag to keep the lantern.

> * Defeating Groal spawns a Covetous Pilgrim in Shakra's Bilewater room.

Added a tag to remove the pilgrim.

> * I had previously reported that Trail's End checks Shakra's vanilla inventory slots instead of checking the map items directly, but while that is true, the `ShopItem` objects themselves check whether the player owns their respective maps in addition to checking whether their slot has been purchased, so the end result is that Trail's End is unlocked by owning all of Shakra's map items. I don't think anyone likes the idea of needing 14 random obtains to unlock Trail's End, a vanilla Silk & Soul requirement, but this implementation is functional.

Out of scope of this PR, it doesn't implement the Trail's End location.  The PR that does that can address changes to Shakra's code as needed.

> 
> *At least, the sitting position is. The bench's interact zone extends outside the trigger, so it's possible to start to sit at it and get interrupted by Shakra. That's a vanilla bug that just never comes up because it's not possible to skip Marrow Shakra, but since it could happen in rando, I suggest editing Mt. Fay's Shakra to just not have that event, or maybe remove it entirely since it has that other problem anyway.

PR #203, opened 2026-04-22
Implement Craw Summons location

### LINKED ISSUE #152: Implement Craw Summons location
This should be an object location replacing the craw summons shiny item, ordinarily spawned randomly at eligible benches. For the ItemChanger implementation, I think additionally the location should allow specifying exactly the list of benches which are eligible, overriding the fsm behavior on all benches accordingly. Also, the location should ensure that the craw summons shiny spawns at every eligible bench (without an rng check, and regardless of whether it has previously spawned at another eligible bench; the Act 3/quest/lever requirements can be kept as is).

Undecided: whether the randomizer should use all benches eligible in the base game, or a single one of these, randomly selected prior to randomization.

thomaslunn:
Adds a module, location, and container for supporting the Craw Summons location.

`DeterministicCrawSummonsModule` specifies which benches Craw Summons can spawn at, and ensures they spawn consistently at these locations. The module can also be used to overwrite the conditions for Craw Summons to spawn. By default, Summons spawn at all vanilla locations (simultaneously) and according to the vanilla conditions.

Resolves #152.


ItemChanger.Silksong/Modules/DeterministicCrawSummonsModule.cs
```diff
@@ -0,0 +1,140 @@
...
+    protected override void DoLoad()
+    {
+        FsmEditGroup editGroup = new();
+
+        foreach (var scene in CRAW_SUMMONS_SCENES)
+        {
+            if (SceneNames.Contains(scene))
+            {
+                ItemChangerHost.Singleton.GameEvents.AddSceneEdit(scene, PatchCrawSummonsAppearedScene);
+                editGroup.Add(new FsmId(scene, "RestBench", "Bench Control"), fsm => ForceSummonsSpawn(fsm, scene));
```
dplochcoder: Consider making a single 'ModifySummonsSpawn' method that checks SceneNames at runtime and handles both cases.  Doing it this way ensures that any late changes (i.e. after Load()) to SceneNames will be respected.

ItemChanger.Silksong/Modules/DeterministicCrawSummonsModule.cs
```diff
@@ -0,0 +1,140 @@
...
+        Benchwarp.Data.SceneNames.Dust_10, // Sinners road broken toll bench
+        Benchwarp.Data.SceneNames.Dust_11, // Sinners road Styx bench
+        Benchwarp.Data.SceneNames.Wisp_04 // Wisp Thicket bench
+    ];
+
+    /// <summary>
+    /// List of scenes that the Craw Summons should spawn at. Defaults to all possible
+    /// vanilla locations simultaneously.
+    /// </summary>
+    public IEnumerable<string> SceneNames { get; init; } = CRAW_SUMMONS_SCENES;
```
dplochcoder: Consider making this a HashSet or List to avoid the extra $type variable in json.

ItemChanger.Silksong/Modules/DeterministicCrawSummonsModule.cs
```diff
@@ -0,0 +1,140 @@
...
+
+/// <summary>
+/// A module which controls the spawning of Craw Summons in act 3.
+/// </summary>
+public class DeterministicCrawSummonsModule : Module
+{
+    /// <summary>
+    /// List of all scenes that support Craw Summons
+    /// </summary>
+    private static readonly string[] CRAW_SUMMONS_SCENES =
```
dplochcoder: nit: IReadOnlyList<string> to make it truly readonly

dplochcoder (changes requested):

ItemChanger.Silksong/Modules/DeterministicCrawSummonsModule.cs
thomaslunn: I actually had this as a list initially, but had to change it due to an odd issue where the SceneNames list would get duplicated copies of all the scene names after each reload - see 40c35409cfc9e20811515f088c561d30fb9c4741. I assume some kind of serialisation problem. Changing to IEnumerable was my fix for it. Do you have any ideas why that might have been happening?

dplochcoder: Ah yes, that's because Newtonsoft's default behavior is to deserialize Lists and Sets into existing containers, not reset them.  So it creates a new DCSM, which already has stuff in it, then reads the SceneNames field and adds all of its stuff into the existing list.

So in general you can't have a non-empty initializer here.  One alternative would be to make this field `required`, so that any caller who instantiates the module has to explicitly set the SceneNames list.

thomaslunn: Makes sense, thanks. I guess required is the way to go then. I'll make the full scene list public in that case as it'll help for initialising the module.

ItemChanger.Silksong/Serialization/QuestCompletionBool.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        get
+        {
+            if (!QuestManager.TryGetFullQuestBase(QuestName, out var quest))
+            {
+                LogWarn($"Unable to locate quest '{quest}'.");
+                return false;
+            }
+
+            return RequiredCompletion switch
```
dpinela: Given that the class itself is already a discriminator, it would make more sense to me to use a distinct class for each completion state. That would also eliminate the possibility of an invalid QuestCompletion value.

flibber-hk: I feel like a single class fits better with the existing style (see BaseGameSavedItem), that said I don’t really know which if any is better

thomaslunn: A version of `QuestCompletionBool` has now been merged, so I've changed this PR to use that.

homothetyhk:
Thanks for the PR! I decided to remove the Appear In Black sequence for the craw pin, since that 7 second wait seems really annoying to go through on every bench. Now it should always first spawn through the regular appear that leaves control.

PR #204, opened 2026-04-25
Greyroot locations

dpinela:
This implements two locations: the Pollip Pouch location and the Cursed Crest one. The Cursed Crest location implicitly triggers a scene change to the Chapel of the Witch (by using the vanilla code path); the items are given when Greyroot snaps Hornet's neck instead of on entry to the chapel.

Both locations are replaced by (separate) shinies in Greyroot's hut once the Rite of Rebirth quest is complete.

This PR includes two things that may be useful elsewhere: a module to suppress giving the Cursed Crest on entry to the Chapel of the Witch, and a boolean value provider that checks whether a quest is completed (used as the condition to switch the two DualLocations introduced here).


thomaslunn:
FYI, `QuestCompleteBool` duplicates functionality from [`QuestCompletionBool`](https://github.com/homothetyhk/ItemChanger.Silksong/pull/203/changes#diff-2bfcec2f5784b81ced7c83367f5cb163493f7f6a5b100a4a6b404aee8ef45db8) over in my pull request #203. I don't mind which one gets implemented, mine can check for more completion states but I'm not using any of it in the PR.

dpinela:
Interesting. Given that the each completion state is represented by a distinct property, I'm inclined to say each of these should be a distinct class entirely instead of using an enum to discriminate (in which case QuestCompleteBool would be one of them). I'll make a note of this in your PR.

ItemChanger.Silksong/Locations/GreyrootCrestLocation.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace ItemChanger.Silksong.Locations;
+
+public class GreyrootCrestLocation : AutoLocation
+{
+    protected override void DoLoad()
+    {
+        Using(new FsmEditGroup()
+        {
+            {new(SceneName!, "Wood Witch", "Dialogue"), HookWitch},
```
flibber-hk: It's pretty likely I've made this mistake in the past, but shouldn't this be UnsafeSceneName?

dpinela: So have I. I tend to forget that property exists. I also find it a little bit of a misnomer; if you care about the value being non-null, throwing an exception on the spot is safer than silently passing on a null value.

ItemChanger.Silksong/Locations/GreyrootCrestLocation.cs
```diff
@@ -0,0 +1,33 @@
...
+        // from a reusable warp.
+        ActiveProfile!.Modules.GetOrAdd<CurseSuppressionModule>();
+    }
+
+    protected override void DoUnload() {}
+
+    private void HookWitch(PlayMakerFSM fsm)
+    {
+        FsmState neckSnapState = fsm.MustGetState("Neck Snap");
+        neckSnapState.AddMethod(GiveAll);
```
flibber-hk: This’ll be easier to test when big ui defs are available, but I have a suspicion this might not properly support big ui defs? (You might need to use AddLambdaMethod or something like that)

dpinela: I don't know if this is because I'm doing something wrong or because there's a bug in the big UI implementation, but I tried substituting AddLambdaMethod here and the FSM still moves on without the big UI being dismissed (using the implementation from #192).

flibber-hk: I had a look at the FSM - I'd guess that the Tk2dWatchAnimationEvents action is what's proccing the FINISHED action? The idea is that the FSM shouldn't be frozen until the UIDef is finished (and it's the location, not the UIDef, which is responsible for taking/giving control). Most likely the simplest fix here is either to move the give to a different state, or add a state between Neck Snap and a different one to hold the give action

dpinela: I'd tried doing exactly what you suggested, and it hadn't worked, but then I updated FsmHelper and AssetHelper like you suggested in the other PR and that made it work.

homothetyhk:
Thanks for the PR!

PR #205, opened 2026-04-26
Util to set starting crest

flibber-hk:
I suggested this in discord and didn't get any pushback


dplochcoder (approved):

ItemChanger.Silksong/Modules/StartCrestReplacementModule.cs
```diff
@@ -63,4 +64,29 @@ bool IsTargetInstruction(Instruction i)
...
 
     private ToolCrest GetStartingCrest() => ToolItemManager.GetCrestByName(ReplacementCrestID);
+
+    /// <summary>
+    /// Set the given crest to be the starting crest, by adding an instance of this module as
+    /// well as using StartLocation and PlayerDataEditModule (editing the CurrentCrestID field).
+    /// </summary>
+    /// <param name="profile">The active profile.</param>
+    /// <param name="crestName">The name of the crest. This should be the item name, not the crest ID.</param>
+    public static void ApplyStartCrest(ItemChangerProfile profile, string crestName)
```
dpinela: The method itself is fine, but I think it makes more sense as an instance method on ItemChangerProfile? Most of what it does doesn't involve this module.

flibber-hk: Makes sense, I put it here mainly because that’s where I’d go looking for it but extension to ItemChangerProfile is probably good (obviously I can’t add it as an instance method…)

Issue #86, opened 2026-03-14
Module to allow returning to Act 2 from Act 3


homothetyhk:
See discussion in the [planning issue](https://github.com/homothetyhk/Silksong-Randomizer-Organization-Stub/issues/5#issue-3827211065). The module should add a lore tablet or similar interactable in Act 3 which, if used, reverts all Act 3 state changes. The module should also override the Benchwarp start location while in Act 3 to be the same as the interactable to exit Act 3. The location of the interactable should be configurable, defaulting to Songclave.

dpinela:
Where should you respawn upon returning to Act 2? I see three reasonable possibilities:

- On the same spot you returned from (where the return interactable was)
- At your last save before fighting GMS
- In the Cradle right next to GMS's arena (the idea being that the return literally "undoes" the effect of that fight)

homothetyhk:
Another option is to respawn at the original start def.

A question for rando is whether the Act 3 -> Act 2 connection needs to be modeled in state logic. On that front, I can't think of any type of state that would be expected to be carried through in base rando, though it is probably good to have the ability to carry state for connections.

In that case, I think last save before GMS is not workable, but the other 3 are all fine. I don't have a strong preference between them. Right next to GMS feels natural for the reason you mention ("undoing" the fight), so maybe we should go with that.

dpinela:
If I understand correctly, last save before GMS would not work because the warp's destination would vary depending on what that last save was, so it would not be possible to carry state, whereas the other options all send you to a fixed location?

homothetyhk:
Right, sending to a fixed location can be represented by a waypoint without any difficulty. Arbitrary respawn is very difficult.

dpinela:
On the technical side, should this be a default module? (What are the criteria for something to be a default module? Have they ever been written down anywhere?)

homothetyhk:
I think there is some disagreement currently over whether where default modules should be on the spectrum from "the bare minimum for components and features provided by IC to function with expected interoperability" to "everything expected in a default randomizer save". With ItemChanger.Core changes, it is no longer possible to create a profile without default modules, so that sort of biases the determination towards bare minimum since it is annoying to remove/add back modules.

On that basis, I'd lean towards this should be nondefault. I think that also makes sense from the perspective that there could be alternative modules that handle it differently, etc, so the choice preferred by the randomizer doesn't necessarily need to be packaged in the host.

PR #206, opened 2026-04-26
Implement Nuu locations

### LINKED ISSUE #89: Implement Nuu locations
Refer to discussion in the [planning issue](https://github.com/homothetyhk/Silksong-Randomizer-Organization-Stub/issues/7#issue-3827225679).

The current proposal is that Nuu locations should require the player to have fought a certain number of bosses, rather than unique enemies. Possibly, the task of counting bosses should be delegated to a module, which provides an API for connections to indicate what counts as a boss or not. Certainly, for the randomizer's purposes, we would like to not count any missable bosses.

The boss thresholds suggested in the linked issue were 10 for the tool pouch and all for the memento. Realistically, "all" is probably not feasible for rando, so it should be a concrete number that can be configured for both locations.

thomaslunn:
This PR contains implementations for both of Nuu's locations:
- `BossKillsCounterModule` counts how many bosses have been defeated by tracking obtained journal entries. The module can also support arbitrary boss definitions. This behaviour is exposed publicly for future connections that may wish to change which bosses contribute.
- `NuuToolPouchLocation` replaces the tool pouch given when completing Bugs of Pharloom wish. The required boss kills count is configurable, defaulting to 10.
- `NuuMementoLocation` replaces Hunter's Memento. The required boss kills count is also configurable, defaulting to 30.
- These locations both have dual `CoordinateLocation`s to handle item persistence.

There are also several utility additions implemented for this feature that may be useful elsewhere:
- Raw data strings have been added for all journal entries. 
- Utility functions have been added for editing quests: `ModifyTargetAmount` can modify counts required for quest completion, ~and `ModifyReward` can overwrite the quest's item reward.~
~- `PlacementSavedItem` is an IC implementation of `SavedItem` based on a `Placement`~

Resolves #89.

~_Note: Currently the boss counting module does not count re-fights for some bosses - I may get around to looking at these soon depending on how important it is._~


thomaslunn:
I'm going to convert this to draft for now:
- There's an issue with Big UI defs not appearing on the tool pouch location
- There is significant overlap with #162 regarding replacing quest rewards

thomaslunn:
This is ready for review now:
- Big UI defs supported for both placements
- Bosses with issues are handled by specialised `BossDefinition`s
- I've taken a different approach to granting the tool pouch placement to avoid dealing with quest rewards

For now all bosses are included, with logic to ensure they are counted as defeated once they get locked out (although this isn't an issue for many bosses due to the planned return-to-act-2 feature, #86).

qwint:
tested:
* multi-item placements, all were granted as expected, but when combined with persistence the chest containers blocked me from mantling when both were previously sent. my shinies happened to fling down, but i can imagine that's fragile and on repeat visits when the chests are already open they are unreachable itemless
* crest uidefs, worked fine on the memento location, but the crafting kit location had the text before hang on screen overlapping with the crest grant ui
* rosary flings, neither flung but granted rosaries without issue
* persistence, besides previously mentioned multi-item issue, honored persistence flag, re-granting when set and not when unset
* previously having full item pouch and the relevant memento, worked without issue

thomaslunn:
Thanks for the testing Qwint! The issues should now be resolved:
- I've relocated the persistence items into locations where being replaced with a chest doesn't cause any navigation issues
- Ended the Nuu dialogue slightly earlier to avoid any overlap with UI defs

homothetyhk:
Thanks for the PR!! I went ahead and added locations for the scroll and journal, and made a few style changes to BossKillsCounterModule.

PR #208, opened 2026-04-29
Add Pin Badge (Pinstress) location

### LINKED ISSUE #140: Implement Pinstress (Pin Badge) location
This is controlled by a FSM on the `NPC` object in Peak_07.

DatUub:
- Adds `PinstressLocation`, an `AutoLocation` that hooks the `NPC / NPC Control / Reward` FSM in `Peak_07` and routes the Pin Badge grant through ItemChanger.
- Mirrors the bell-hermit / yarnaby pattern: strips `GetQuestReward` and `SavedItemGetV2` from the Reward state and inserts `GiveAll`. `EndDialogue` and the quest state writes (`SetPlayerDataVariable`, the `EndQuest` in the next state) stay intact so the conversation closes and the quest resolves normally.
- Registers `Pin_Badge` in `BaseLocationList` pointing at the new location.
- Adds a `PinstressLocationTest` under `LocationTests` that places `Surgeon's Key` and `Flea` at the slot and starts near `Peak_07`.
- [ ] Build solution
- [ ] Run `Pinstress Location` test in-game: defeat Pinstress, return and talk to her, verify both items grant via IC and the dialogue/quest closes correctly.
- [ ] Confirm vanilla Pin Badge no longer appears (no double-grant).
Closes #140.


qwint:
(merged conflicts, they seemed to be simple "multiple inserts in the same place" issues)
tested:
* [issue]: crest itemdefs, the previous text box does not get cleared before the crest ui
* lore itemdefs, worked without issue
* rosary item flings, did not fling but granted rosaries without issue
* persistence, granted items with persistence flag on re-fight, did not grant items without persistence flag on re-fight, but did allow re-fight regardless
* tested with previously receiving pin badge, worked without issue


One thing of note, this seems to be forced by the pinstress fight (tested vanilla wings/act3 and this didn't happen), but the path to and up from pinstress to the top exit are both missing one of the platforming-enemies making it not wingless movement.

qwint:
retested, crests now display as expected other test cases unchanged,

I still think I would prefer a shiny/chest for the persistence instead of a permanent refight, and am worried about the scene changes affecting traversal logic but I'll leave those final says to homothety

PR #209, opened 2026-04-30
Add Grounded Sprint novelty item (#52)

### LINKED ISSUE #52: Novelty item: grounded sprint
Premise: Silksong is very large, and a randomizer with late Swift Step could be very unpleasant. On the other hand, starting with Swift Step or forcing Swift Step to appear early is not desirable since Swift Step is an important progression item.

I propose an item which:
- While Hornet is grounded, holding the SPRINT button allows her to sprint at the same speed as if she had Swift Step.
- While Hornet is not grounded, her speed is capped to walking speed.
This could be given as a starting item or randomized progressive with Swift Step. The main goal is that capping midair speed removes most of the logic implications of this item, so that its main effect is QoL/balancing for races.

DatUub:
Closes #52.

Grounded Sprint: hold dash on the ground to sprint with **vanilla sprint animations and cState**, without midair sprint or dash for logic.

## Approach (flibber review)

**Not** a `GetRunSpeed` speed hack. The sprintFSM's `PlayerDataBoolTest(hasDash)` gate is rewritten (named states first; discovery logs extras) to:

`hasDash || (hasGroundedSprint && onGround)`

so the vanilla FSM owns animations, Sprintmaster / Quickening tiers, and cState. Direct field `playerData.hasDash` stays false until real Swift Step (`CanDash`, progression).

### Extra guards (issue #52 air policy)
- Air: `CANCEL SPRINT` + clear `sprintBufferSteps`
- Ground sprint: `PreventShuttlecock()`
- `HeroJump(bool)` prefix blocks shuttle-cock when only Grounded Sprint is owned

## Test plan
- [ ] Hold dash on ground: sprint **animation** + Swift Step speed
- [ ] Leave ground while sprinting: cancels, walk speed, no shuttle-cock
- [ ] No air dash
- [ ] Full Swift Step restores vanilla
- [ ] Sprintmaster / Quickening tiers apply

FSM viewer: HeroController → sprintFSM → states with `PlayerDataBoolTest` / `hasDash` (see `KnownHasDashGateStates` + log line on first patch).

### DIFF
--- a/ItemChanger.Silksong/Modules/CustomSkills/GroundedSprintModule.cs
+++ b/ItemChanger.Silksong/Modules/CustomSkills/GroundedSprintModule.cs
@@ -11,10 +11,13 @@ namespace ItemChanger.Silksong.Modules.CustomSkills;
 /// flibber (#209): trick the game so vanilla sprint anim/cState run.
 ///
 /// Playtest fixes:
-/// - Ledge fall: hard cancel + clamp air horizontal speed to walk.
+/// - Ledge fall: soft cancel + clamp air horizontal speed to walk.
 /// - Wall clip: never allow real HeroDash for GS-only.
 /// - Ground dash still worked: Prepatcher makes playerData.hasDash use GetBool, so
 ///   spoofing hasDash made CanDash true. Force CanDash false and block HeroDash*.
+/// - Jump-from-sprint / downslash: never send HARD LANDING (kills jump Y and interrupts
+///   air attacks). Only CANCEL SPRINT + flag clear, and only on leave-ground / jump —
+///   not every airborne frame.
 /// </summary>
 public class GroundedSprintModule : CustomSkillModule
 {
@@ -237,12 +240,14 @@ public class GroundedSprintModule : CustomSkillModule
     private static void HeroJumpBoolPrefix(HeroController __instance, ref bool checkSprint)
     {
         if (!OnlyGroundedSprintKit()) return;
+        // Disable shuttlecock path without wrecking jump velocity.
         checkSprint = false;
         __instance.PreventShuttlecock();
         SprintBufferStepsField?.SetValue(__instance, 0);
         SyncBufferStepsField?.SetValue(__instance, false);
         NoShuttlecockTimeField?.SetValue(__instance, Time.timeAsDouble + 5.0);
-        CancelSprintHard(__instance, clampAirSpeed: false);
+        // Soft cancel only — HARD LANDING / air clamp here was killing jump height.
+        SoftCancelSprint(__instance, sendFsmEvent: true, clampAirSpeed: false);
     }
 
     /// <summary>Before LeftGround fills sprintBufferSteps from isSprinting/dashing.</summary>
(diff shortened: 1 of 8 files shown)


ItemChanger.Silksong/Modules/CustomSkills/GroundedSprintModule.cs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private void EnsureCachedSpeeds(HeroController self)
+    {
+        if (cachedBase != null) return;
+        if (self.sprintFSM == null) return;
+
+        try
+        {
+            foreach (var state in self.sprintFSM.FsmStates)
```
flibber-hk: Rather than iterate over the fsm states, it would be better for maintainability (and performance) if you just looked at the correct state that has the action. If it's not in the obvious place you might want to leave a comment indicating where the fsm is (so people can find it in FSM viewer).

ItemChanger.Silksong/Modules/CustomSkills/GroundedSprintModule.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                foreach (var action in state.Actions)
+                {
+                    if (action.GetType().Name != "ConvertDoubleBoolToFloat") continue;
+
+                    var t = action.GetType();
```
flibber-hk: Doing this reflection is bizarre; it would make more sense to just write C# code to get the fields from the action of whatever type it is (e.g. using a method like GetActionOfType\<whatever\>)

flibber-hk:
As I indicated on the previous PR, my instinct is that this would make more sense by tricking the game into thinking the player has dash at certain times rather than modifying the speed (so animations, cstate etc work, as you noted). I don't want to say that my proposed implementation strategy is better (or worse) than yours a priori - obviously you've looked at the relevant code more than me - but if you think this is the correct approach, I think this PR ought to include a version where the Hornet running animation works properly (etc).

(EDIT - to be clear, I'm not objecting to submitting a partially-implemented feature, but rather that it's not clear to me that this approach can extend naturally to the complete feature and the simplest way to prove that it can is to do it :zote:)

(I haven't actually tested this, I'm just going by the fact that you don't appear to have made any changes that address the animations issue and as you say they haven't been done)

DatUub:
Rewrote on top of current master addressing the review feedback.

Instead of overriding run speed, Grounded Sprint now makes `PlayerData.GetBool("hasDash")` true only while grounded (via CustomSkillPlayerDataModule). The vanilla sprint FSM owns animations, cState, and Sprintmaster/Quickening tiers. Direct `hasDash` field reads stay false so dash/progression stay gated.

Also cancels sprint in air, blocks shuttle-cock, and clears the leave-ground sprint buffer.

Ready for re-review when you have a chance.

DatUub:
Redid this the way you suggested: no more run-speed spoof.

The sprintFSM's hasDash PlayerDataBoolTest is rewritten to true when the player has Grounded Sprint and is on the ground (named states first, with a one-time discovery log if the list is incomplete). Vanilla FSM handles anim/cState/speed tiers. Field hasDash stays false so dash/logic stay gated; air cancel + PreventShuttlecock + HeroJump guard keep midair clean.

Happy for another look when you have time.

PR #212, opened 2026-05-03
Add Creige Crafting Kit location

### LINKED ISSUE #136: Implement Creige location
This is controlled by a FSM on `HH Bartender` in Halfway_01.

DatUub:
- Adds `CreigeLocation`, an `AutoLocation` that hooks the `HH Bartender / Dialogue` FSM in `Halfway_01` and routes the Crafting Kit grant from the Crawbug Clearing wish through ItemChanger.
- Strips `GetQuestReward` and `SavedItemGet` from the `Quest Reward` state, ends conversation, then `GiveAll` (crest UI no longer stacks on NPC text).
- Persistence / re-give hooks **`General Convos` only** (after quest COMPLETE). Does **not** grant from `Quest State` (that path also runs before Ragpelt turn-in).
- Registers `Crafting_Kit__Creige` in `BaseLocationList` pointing at the new location.
- Adds a `Creige Location` test under `LocationTests` that places `Surgeon's Key` and `Flea` at the slot, spawns near Creige in `Halfway_01`, and marks the Crawbug Clearing quest ready-to-complete.

- [x] Build solution
- [x] Address review: crest text-box clear before GiveAll
- [x] Address review: persistence without early grant on first dialogue
- [x] Confirm vanilla Crafting Kit no longer appears (no double-grant)

Closes #136.

### DIFF
--- a/ItemChanger.Silksong/Locations/CreigeLocation.cs
+++ b/ItemChanger.Silksong/Locations/CreigeLocation.cs
@@ -1,0 +1,68 @@
+using ItemChanger.Locations;
+using HutongGames.PlayMaker;
+using HutongGames.PlayMaker.Actions;
+using QuestPlaymakerActions;
+using Silksong.FsmUtil;
+
+namespace ItemChanger.Silksong.Locations;
+
+/// <summary>
+/// Location for the Crafting Kit granted by Creige (HH Bartender) in Halfway_01
+/// after Hornet completes the Crawbug Clearing wish. Strips the GetQuestReward
+/// and SavedItemGet actions from the Dialogue FSM's "Quest Reward" state and
+/// routes the grant through the IC placement. The preceding "Quest Complete
+/// Prompt" state still runs EndQuest, so the wish closes normally.
+/// </summary>
+public class CreigeLocation : AutoLocation
+{
+    protected override void DoLoad()
+    {
+        Using(new FsmEditGroup()
+        {
+            { new(SceneName!, "HH Bartender", "Dialogue"), HookCreige },
+        });
+    }
+
+    protected override void DoUnload() { }
+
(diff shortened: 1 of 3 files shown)


qwint:
tested
* [issue]: crests did not clear the creige text box before showing the crest ui
* lore showed text as expected
* rosaries _did not fling_ but were granted as expected
* [issue]: persistent tags were not respected
* full tool upgrades/pouch did not affect location
* multi-item placements worked as expected (was in the test already)
* tested interactions with the nectar, was prompted to turn in pelts with both ready and then could handle nectar, was prompted to get nectar when not all pelts were gathered. I presume this is vanilla handling but wanted to test regardless.

qwint:
the recent updates seem to have forced item granting onto the first dialogue message for the quest,
meaning there's no more dialogue overlap, you can get the item without turning in the pelts, and potentially related but lore uidefs seem to no longer show up at all

I did see persistence honored, but it seemed to just grant the item at the start of any dialogue tree after you got the item once, potentially just affected by the same issue as the first item but regardless seemed strange

DatUub:
shall take a look thank u for the feedback

DatUub:
Addressed the early-grant regression from the persistence change:

**Cause:** the re-give hook was on `Quest State`, which also runs before Ragpelt turn-in (NONE / ACTIVE / Can Complete?), so items could grant on the first dialogue line.

**Fix:** first grant stays on `Quest Reward` (`EndConversation` then `GiveAll` for crest UI). Persistence re-give is only on `General Convos` (post-COMPLETE path).

Also rebased onto current `master`.

Local checks: build clean; in-game LocationTest load to Halfway_01; FSM drive scenarios for no early grant / reward grant / persistent re-give.

PR #218, opened 2026-05-08
Implement the Rune Rage location

dplochcoder:


qwint:
tested:
* rosary flings, did not fling but granted rosaries without issue
* persistence, did honor persistence flag, re-granting items if set and not re-granting if unset
  * from a design standpoint though I personally don't like how you 1) can always re-challenge the fight even if there is no persistence reward 2) the re-challange refreshes on room load including finishing the fight, so pressing up after getting the first reward sends you directly into a re-fight
* tested the location while already having rune rage in inventory, worked without issue

dplochcoder:
> tested:
> 
> * multi-item placements, worked without issue
> * rosary flings, did not fling but granted rosaries without issue
> * persistence, did honor persistence flag, re-granting items if set and not re-granting if unset
>   
>   * from a design standpoint though I personally don't like how you 1) can always re-challenge the fight even if there is no persistence reward 2) the re-challange refreshes on room load including finishing the fight, so pressing up after getting the first reward sends you directly into a re-fight
> * tested the location while already having rune rage in inventory, worked without issue

This was fun, there were several issues to fix here:
 - `PlayerData.defeatedFirstWeaver` is set in the memory scene, so it wasn't being set at all (fixed), but
 - The shrine checks for `PlayerData.hasSilkBomb` instead. so changed that check, but
 - Because this was implemented as a DualLocation, the 'fix' code doesn't run after the weaver is defeated, so changed this to a top-level AutoLocation with manual coordinate spawning when the boss is defeated.
 
 This should fix all the re-challenge issues.

qwint:
tested all:
* rosary flings, did not fling, but no issues granting rosaries
* tested while already having rune rage, worked without issue

the statue now breaks as expected and you cannot refight, and persistence is instead a shiny/chest on the ground which works great
I am still seeing though that the scene you wake up from after the fight already has the persisted shiny/chest, unsure if that's something you want to fix

dplochcoder:
> tested all:
> 
> * multi-item placements, worked without issue
> * rosary flings, did not fling, but no issues granting rosaries
> * persistence, honored persistence flag, both when set and unset
> * tested while already having rune rage, worked without issue
> 
> the statue now breaks as expected and you cannot refight, and persistence is instead a shiny/chest on the ground which works great I am still seeing though that the scene you wake up from after the fight already has the persisted shiny/chest, unsure if that's something you want to fix

Changed the respawn logic so this should be fixed now.

qwint:
retested and i am still seeing the persistence shiny/chest in the scene after you bind first sinner (and get the item for the first time)

dplochcoder:
> retested and i am still seeing the persistence shiny/chest in the scene after you bind first sinner (and get the item for the first time)

Take three; turns out I was incorrect about the order of item grant / scene reload.  I also didn't realize we supported fully persistent items (I'm not sure those will exist outside of testing though, most should be semi-persistent?)

homothetyhk:
Yeah, Persistent is mainly for ease of testing and also to set the boundary on how often items can be refreshed (so, for example, location implementations can assume that they do not have to give the same item more than once per scene). The only example in HK.IC is Leg Eater repair items that may or may not be used anywhere (arguably, Soul_Totem-Path_of_Pain should be Persistent instead of SemiPersistent). In IC.SS, I could imagine things like flea brew refills or plasmium refills being Persistent (some enemy drop quest items might also make sense as persistent on connections, and plandos might want persistent shell shards, etc).

homothetyhk:
I removed the last few lines of the after-fight fsm edits; they seemed irrelevant, since `skipState` has no outgoing transitions.

PR #220, opened 2026-05-09
Implement SceneEditGroups.

dplochcoder:
Splitting this out to its own PR for simplicity.


flibber-hk:
This one probably should be reconciled with DisposableSceneEdit which I added as part of my PR (I didn’t do a group because I only needed one scene edit)

dplochcoder:
> This one probably should be reconciled with DisposableSceneEdit which I added as part of my PR (I didn’t do a group because I only needed one scene edit)

Done.

flibber-hk:
(Fwiw the answer may actually be that this should be done by IC.Core… obviously it would be fine to leave that as a “BadMagic can do that if he wants” though)

PR #223, opened 2026-05-10
Make the Twisted Bud cry when spawned as a Shiny

dplochcoder:


ItemChanger.Silksong/Items/ItemChangerSavedItem.cs
```diff
@@ -49,17 +50,19 @@ public override bool Redundant()
     /// <param name="playerDataBoolName">A supplementary PlayerData bool, used by some items.</param>
     /// <param name="customUIDef">Function used to generate the UIDef. The default UIDef will
     /// be passed as the argument.</param>
+    /// <param name="tags">Custom tags to add to the item.</param>
```
flibber-hk: Is there any reason why you need to pass the tags here rather than using AddTag or (more likely) WithTag to add them? Pre-empting the follow up question, the uidef annoyingly has to be done like this because it’s init-only so I can’t modify it after the item has been constructed.

dplochcoder: Not particularly, just wasn't aware of that set of extensions.

ItemChanger.Silksong/Tags/SpecialItemTags/TwistedBudCryTag.cs
```diff
@@ -0,0 +1,97 @@
...
+using UnityEngine;
+
+namespace ItemChanger.Silksong.Tags.SpecialItemTags;
+
+public class TwistedBudCryTag : ShinyModifierTag
+{
+    public override void ModifyShinyContainer(Placement placement, Item item, GameObject shiny)
+    {
+        // Don't modify the vanilla location.
+        // TODO: The location should also disable the Crying Audio Control object when randomized, unless it's a dearest.
```
flibber-hk: Style question: does it make more sense to simply disable the crying audio unconditionally, and modify the vanilla location if appropriate? This would be simpler code-wise, I imagine. My instinct is that it would be idiomatic to do this but I can’t think of an example from HK so perhaps I’d defer to Homothety

dplochcoder: > Style question: does it make more sense to simply disable the crying audio unconditionally, and modify the vanilla location if appropriate? This would be simpler code-wise, I imagine. My instinct is that it would be idiomatic to do this but I can’t think of an example from HK so perhaps I’d defer to Homothety

I would say definitely yes if the tag reproduced the vanilla behavior exactly, but it doesn't here for the commented reasons so it's more of a gray area.  I'll defer either way here.

PR #226, opened 2026-05-12
Implement Styx cocoons and Skynx shop

### LINKED ISSUE #170: Implement Styx Shop
Been thinking about this in the context of https://github.com/homothetyhk/ItemChanger.Silksong/issues/85.  It's probably a low-priority feature for now but I'd like to record some thoughts for future work.  Egg Shop is good because it gives value to an item that seasoned players otherwise find worthless, and because it makes early Simple Key decisions interesting; for similar reasons, I imagine Styx Shop could be good as well.

The way I'd implement this would be:
 * Enabling Styx shop and randomizing Silkeaters are independent toggles
 * If Styx shop is enabled:
   * It presents similarly as Egg Shop.  Talking to Styx after introductions shows a list of items for sale with cumulative Silkeater costs
   * If the player consumes any Silkeaters for their vanilla purpose, _those_ Silkeaters can grow back at an accelerated rate to be farmed, but no additional ones will grow.  (The base game growth rate is 1 per hour; for rando we should make it something like 5-10min.)  In this way, the player is never locked out of Styx shop rewards no matter how many Silkeaters they consume, nor can they farm additional Silkeaters from the nest to go out of logic.
   * Once the player has bought out Styx shop completely, the cap on farming is removed but the accelerated rate is also reverted back to normal.
 * If Silkeaters are randomized, the 9 vanilla Silkeaters in the world are randomized as well as two more in Styx's nest.
   * The first Silkeater produced, and the first Silkeater produced after completing the Queen's Egg courier quest, are both replaced with randomized items.  The cocoons for the randomized items grow immediately and do not require the player to wait at all for Styx to grow them.  If the player has collected neither before completing the courier quest, the two randomized items are given sequentially from the next two silkeater spawns.
   * All further Silkeaters produced are vanilla, if their growth is allowed by Styx shop settings.

Alternatives to consider:
 * Do not allow the player to consume Silkeaters at all until the shop is bought out.  (I think this is probably too harsh, we should expect players to still be dying frequently for a while.)
 * Do not randomize any grown Silkeaters, only the 9 placed ones.
 * Do not implement Styx shop at all (boo)

dplochcoder:
This theoretically closes #170 and lays a bunch of other useful groundwork.  This is a pretty opinionated addition so I fully expect some pushback and changes, but my hope is that the general idea will still stick.

- Implements SilkGrubCocoons as a generic container.  These can be used to preserve the containers for Silkeaters in the wild, when those locations are randomized.
- Coordinates Styx and Skynx under a new StyxAndSkynxModule, which ensures both NPCs are accessible, implements Skynx shop almost identically to Egg shop, and controls Styx's cocoon-growing behaviors as outlined in the linked issue.


qwint:
tested:
* multi-item placements, worked without issue; multiple shinies spawned from styx and extra items added to the skynx shop
* rosary flings, styx's shinies fling rosaries and skynx flings rosaries directly
* tested the location while already having max silkeaters, worked without issue

additional testing notes:
* [issue]: skynx will still fling shinies but without spending silkeaters if you choose no
* styx cocoons sometimes fling from the wrong side of the web when broken

dplochcoder:
> tested:
> 
> * multi-item placements, worked without issue; multiple shinies spawned from styx and extra items added to the skynx shop
> * rosary flings, styx's shinies fling rosaries and skynx flings rosaries directly
> * persistence, honored persistence flag, both when set and unset
> * tested the location while already having max silkeaters, worked without issue
> 
> additional testing notes:
> 
> * [issue]: skynx will still fling shinies but without spending silkeaters if you choose no

Fixed this; changed the conversation flow into two consecutive prompts, so the player can decide whether (a) to spend silkeaters, if no (b) to receive paid-for items.  Fixed oversight in FSM editing that purchased items even when player said no.

> * styx cocoons sometimes fling from the wrong side of the web when broken

qwint:
tested all:
* rosary flings, silkeaters flung shinies that then flung rosaries, and skynx flung rosaries directly, but no issues granting rosaries
* tested while already having all silkeaters, worked without issue

retested skynx dialogue paths, saying no to spending silkeaters works as expected, and saying yes/no to the received pre-payed items works as expected

retested cocoons flinging from the wrong side of styx when broken and could still reproduce. wasn't sure from your comment if you meant to address that? but in case reproduction steps help:
* load the styx test
* grant a queen egg
* save styx, hand in the queen egg
* reload the room
* break the left coccoon, it flings from the right cocoon instead

but since this is purely visual I personally wouldn't consider this blocking and am happy with the current implementation

dplochcoder:
> but since this is purely visual I personally wouldn't consider this blocking and am happy with the current implementation

Fixed; it's a pretty jarring bug nevertheless, even if it's only visual.

qwint:
just retested the reproduction steps I listed previously, the shinies now fling from the correct cocoon locations

PR #227, opened 2026-05-12
Implement Clock module.

dplochcoder:
This lays some ground work for <https://github.com/homothetyhk/ItemChanger.Silksong/pull/226>.  It implements a generic clock module that can be used to measure elapsed game time over extended periods.


homothetyhk:
Is there any reason to prefer a harmony patch over a custom monobehaviour on a dontdestroyonload object?

dplochcoder:
> Is there any reason to prefer a harmony patch over a custom monobehaviour on a dontdestroyonload object?

Uh, not really.  That would probably be simpler than this.

Issue #201, opened 2026-04-20
Flea-count location/placements


flibber-hk:
The scope of this issue definitely includes locations corresponding to the following items:

* Mooshka’s Silk Spool
* Mooshka’s tool pouch
* Egg of Flealia

It may or may not include Flea Brew, depending on how the issue is implemented.

I think this could be done in a couple of ways:

* Randomizing the items given by Mooshka directly (so you get the randomized item in the way you’d normally get it - depending on whether or not you travel with the caravan), perhaps with some special handling for re-gives (talking to mooshka if the caravan is still at the location, maybe a shiny on the ground if the caravan has moved or alternatively talking to mooshka at his new location)
* Cost chest placement, where the chest travels with the caravan. This might be nice to have as a novelty setting, if not default.
* A custom shop. This is _probably_ blocked by #178 - custom shops are certainly something we want in itemchanger eventually but that doesn’t mean it should be how mooshka is done.

Caravan re-use is not in scope for _this_ issue, but the fact that we’re going to have to do it anyway means that we don’t have to care about the logical implications for option 2. (I have a bunch of work on the caravan re-use point so it’s probably not worth doing that right now)

dplochcoder:
> (I have a bunch of work on the caravan re-use point so it’s probably not worth doing that right now)

I imagine this will affect https://github.com/homothetyhk/ItemChanger.Silksong/pull/199/changes#diff-269e6c4e97ba0758bc7a0fbe6451641cfaafc5e50cf5ad45cf280278ec949f69 to some extent

PR #228, opened 2026-05-12
Implement custom shops

dplochcoder:
This PR lays the groundwork for implementing custom shops on an existing NPC.  Originally this was going to be part of <https://github.com/homothetyhk/ItemChanger.Silksong/pull/226>, but after some reflection I decided on a different approach, so this is now split out separately.

<https://github.com/homothetyhk/ItemChanger.Silksong/issues/201#issue-4294401777> mentioned custom shops, so linking that here.


ItemChanger.Silksong/Assets/AssetCache.cs
```diff
@@ -57,6 +57,10 @@ public static T GetAsset<T>(this string key)
 
         return GetObjectCache<T>().GetAsset(key);
     }
+    /// <summary>
+    /// Get the source prefab for a game object asset, without instantiating it. Should not generally be used, only when you are certain instantiation isn't necessary.
+    /// </summary>
+    public static GameObject GetPrefabAsset(this string key) => GetObjectCache<GameObject>().GetAsset(key);
```
flibber-hk: I believe this function exists already in this file (line 86)

dplochcoder: I'd like to believe it didn't exist when I originally wrote this function but it's possible I was blind.

flibber-hk: That’s probably true, I added it in my pr from a few weeks ago

dplochcoder:
Added a test, using Mask Maker as an example.

PR #231, opened 2026-05-18
Add a Harmony util for modules with patches

dplochcoder:


homothetyhk:
I think I'd like for the name to change to `UsingPatches` to more clearly relate to `Using`, and for the XML comment to note that disposal in Unload is handled.

Alternatively, maybe it could make sense to do as a `HarmonyEditGroup` instead of an extension. The edit group pattern is a little overengineered for this use case, since most modules will only have a single patch class, but it might be preferable for consistency. Plus, it avoids having a public extension on Module cluttering the interface for module consumers (though I don't think it's a huge issue). Either way is fine by me.

dplochcoder:
> I think I'd like for the name to change to `UsingPatches` to more clearly relate to `Using`, and for the XML comment to note that disposal in Unload is handled.
> 
> Alternatively, maybe it could make sense to do as a `HarmonyEditGroup` instead of an extension. The edit group pattern is a little overengineered for this use case, since most modules will only have a single patch class, but it's might be preferable for consistency. Plus, it avoids having a public extension on Module cluttering the interface for module consumers (though I don't think it's a huge issue). Either way is fine by me.

How about `HarmonyPatchGroup`?

homothetyhk:
Sure, that sounds good.

PR #234, opened 2026-05-22
Implement Silkeater Locations

dplochcoder:
This is trivial to do after implementing the SilkGrubCocoonContainer in https://github.com/homothetyhk/ItemChanger.Silksong/pull/226, so did that here.  To keep things simple, force-spawned the Silkeater in Blasted Steps even in Steelsoul mode.

Also curiously one Silkeater (Terminus) was missing from LocationNames; added that in.


qwint:
this does not ship with a silkeater test, and the one I made to test it working with different item placements I couldn't even get a Surgeon's key shiny to spawn (I didn't test exhaustively, but checked exhaust organ and blasted)
(also confirmed my test worked fine when targeting `Rosary_Chest__Sinner_s_Road`)

dplochcoder:
> this does not ship with a silkeater test, and the one I made to test it working with different item placements I couldn't even get a Surgeon's key shiny to spawn (I didn't test exhaustively, but checked exhaust organ and blasted) (also confirmed my test worked fine when targeting `Rosary_Chest__Sinner_s_Road`)

Fixed this; wound up being a scene initialization order issue that happened on wild cocoons but not the ones in Styx's room because those get manipulated a frame later.

qwint:
tested all (but the queen egg location in the styx pr):
* rosary flings, silkeater flung a shiny which then flung rosaries when picked up, but no issues granting rosaries
* tested a subset of locations while already having max silkeaters, worked without issue

dplochcoder:
> * rosary flings, silkeater flung a shiny which then flung rosaries when picked up, but no issues granting rosaries

Changed CurrencyItem to fling from SilkGrubCocoon containers.

qwint:
just retested rosary flings in single-item and multi-item placements (and persistence) and worked as expected,
rosaries flung from cocoon as expected in both placement cases, and were granted by a preexisting shiny when rechecked and persistence was on

PR #238, opened 2026-05-27
Fix rosary spawn bugs.

dplochcoder:
This removes the cap on currency item spawns to prevent scams.

While not necessary, this also removes the large smooth geo prefab from rosary distribution.  The fix for the above makes them now spawn correctly, but they all render in front of the other currency which looks weird.


homothetyhk:
Looks good. I'm going to make it a default module, since I don't like the idea of unrelated behavior changes based on whether geo chests are randomized or not.

PR #248, opened 2026-07-09
CrestSlotUnlockItem and LockedCrestPreviewModule

### LINKED ISSUE #22: Module to display state of locked crests
To support the solution for randomized memory crests, it should be possible to view what tool slots are available even if the crest is not yet available. I think this module is probably not a default module but rather one that is added on load by `CrestSlotUnlockItem`s. The module should:
* Clearly indicate that locked crests are locked
* Show only unlocked crests by default
  * A method should be available for items to make a crest visible
  * Ideally, a global setting is available for players to make all locked crests visible by default if desired
* Special consideration may be required for hunter crest evolutions, both because I'm given to understand they are treated internally as different crests and because in a world where evolutions are not progressive with the crest itself (which someone may do someday) it would be desired to see the evolution state even if the crest itself is not available - this is negotiable if TC's implementation would make such a state nontrivial to achieve without additional modifications

### LINKED ISSUE #23: CrestSlotUnlockItem
To make memory lockets randomization-safe, they will be replaced with items which unlock specific slots in specific crests, e.g. "Hunter Red Slot Unlock". This story is to implement such items.

* The UIDef for the item should be a corner message which shows the icon of the slot type unlocked
* The implementation should ideally be reasonably extensible to custom crests and tool types as supported by needlework or similar mods for connections to use
* Obtaining a slot for a locked crest should ensure the crest is displayed in #22 
* Obtaining the item should immediately unlock the slot (no need for applying the upgrade such as with a memory locket)
* Some consideration will be needed for crests that allow multiple unlocks of the same type

SpaceMonkeyy86:
Adds a new type of item that unlocks an additional tool slot of a specific type and crest when collected. This is meant as an optional (but likely enabled by default) replacement to memory locket randomization as outlined in https://github.com/homothetyhk/Silksong-Randomizer-Organization-Stub/issues/8.

Since this item allows unlocking extra slots on crests that may not be unlocked themselves, LockedCrestPreviewModule has been added to allow the user to tell that this has happened. Locked crests are shown alongside other crests in the crest switching menu, but they are grayed out and have "(Locked)" appended to their name. By default, only locked crests which have extra slots collected are shown, but there is a setting to show all locked crests instead. There is also a method to make a specific locked crest visible, which is used by CrestSlotUnlockItem when collected.

Implementation details:
- A CrestSlotUnlockItem does not track a specific individual slot on the crest, rather it just tracks the slot type. Instead, when collected, it finds the first locked slot of that type and unlocks it. This simplifies many parts of the implementation.
- LockedCrestPreviewModule is not included by default and is only added if a CrestSlotUnlockItem is loaded. Its effects will only be visible if a CrestSlotUnlockItem is collected for a crest which has not yet been unlocked.
- The sprites for each type of slot are included in the mod files. While the sprites do exist in the base game in theory, they are grayscale and require the color field on the SpriteRenderer to be set to show properly. This does not work well with the current system for displaying base game sprites, so I colored the images manually.
- CrestSlotUnlockItem should support custom crests without issue. It does not support custom tool types.
- There is no special handling for the different hunter crest tiers. This only matters if a connection allows unlocking later tiers without having earlier tiers unlocked, and that seems like a fairly pointless change that would break many assumptions throughout the game's code.

Closes #22 
Closes #23


SpaceMonkeyy86:
@pyuk-bot The issue was due to a crash from trying to access the ToolItemManager class before it had been set up. Not sure how that was working for me before, but it should be fixed now.

